PR #24, opened 2026-04-19
Annotation propagation infrastructure for sibling annotations on @KsMethod

### LINKED ISSUE #11: Annotation propagation infrastructure for sibling annotations on @KsMethod
## Motivation

Multiple planned features want to attach method-level metadata beyond the `@KsMethod` endpoint string: notification vs request (#TBD), typed error payloads (#TBD), per-method timeouts (#TBD). Each wants a sibling annotation whose value travels through the compiler plugin into the generated `RpcMethod` descriptor, so that protocol-specific serialization layers (jsonrpc, HTTP, etc.) can read the relevant bits at the right moment.

Rather than bolt on bespoke handling for each, this issue tracks the shared infrastructure: a general mechanism for declaring and propagating sibling annotations on `@KsMethod` members.

## Proposed shape

1. A convention or marker (e.g. meta-annotation) that identifies an annotation as "RPC-relevant metadata" to be captured by the compiler plugin.
2. The compiler plugin reads any such annotation on a `@KsMethod` function and stores its arguments in the generated `RpcMethod` descriptor (or a metadata map attached to it).
3. Protocol layers (ksrpc-jsonrpc, ksrpc-ktor, etc.) read the descriptors they care about when serializing a call or a response. Annotations a given transport doesn't understand are ignored.

## Dependent issues

This infrastructure is a prerequisite for:

- `@KsNotification` — explicit opt-in for notifications (see sibling issue)
- `@KsErrorData` + `KsrpcException` — typed error payloads (see sibling issue)
- `@KsTimeout` — per-method timeout (see sibling issue)

Each of those issues can be implemented after (or alongside) this infrastructure lands.

## Open design decisions

- Is the metadata a typed map in `RpcMethod`, or does each annotation get a generated accessor?
- Do protocol layers query by annotation type, or does the core publish a merged view?
- How do we handle unknown annotations in older ksrpc versions reading newer service declarations (forward compatibility)?

Implementer's call.

## Motivating use case

Blocks LSP client work in `Monkopedia/kodemirror` — every annotation-based feature below needs this mechanism.

Monkopedia:
## Summary

- Introduces `@KsMethodMetadata`, a meta-annotation that opts a sibling annotation in to compiler-driven propagation onto `@KsMethod` functions.
- Compiler plugin captures qualifying siblings, decodes their arguments into a sealed `MetadataValue` hierarchy, and attaches the result to the generated `RpcMethod` as `List<MethodMetadata>`.
- Test-only demo annotations (`@KsTestMarker`, `@KsTestFlag`) prove end-to-end propagation from source annotation through FIR/IR generation to the runtime descriptor.

## Approach

**Marker mechanism**: a meta-annotation (`@KsMethodMetadata`) applied to annotation classes. Simpler than per-annotation registration and keeps the set of captured annotations open-ended.

**Descriptor shape**: generic `List<MethodMetadata>` rather than per-annotation accessors. Each entry carries the annotation's FQ name and a `List<Pair<String, MetadataValue>>` of captured arguments. Consumers look metadata up by FQ name via `rpcMethod.metadata("com.example.MyMarker")`.

**Typed values, not `Any?`**: `MetadataValue` is a sealed hierarchy — `StringValue`, `IntValue`, `LongValue`, `BooleanValue`, `DoubleValue`, `FloatValue`, `KClassValue` (by FQ name), `EnumValue` (by FQ + entry name), and `ListValue`. Matches the "use types from ksrpc-core" directive. Nested annotation values are explicitly unsupported and fail at compile time with a clear message.

**Forward compatibility**:
- `RpcMethod`'s new `metadata` parameter uses `@JvmOverloads`, preserving the original four-arg constructor's binary signature.
- The compiler plugin references the new types via `maybeReferenceClass` (soft lookups). When it resolves against an older ksrpc-core that predates this change, it quietly falls back to emitting the four-arg constructor call — the plugin still works against old runtimes.
- `MethodMetadata` identifies annotations only by FQ name and stores arguments generically, so an older ksrpc reading a descriptor produced by a newer one can still inspect entries for annotations it doesn't have on its classpath.

## Test vehicle

`ksrpc-test/src/commonTest/kotlin/MethodMetadataPropagationTest.kt` introduces two demo sibling annotations — `@KsTestMarker(tag, priority, flags)` and `@KsTestFlag(enabled)` — plus a negative-control `@NotMetadata` that lacks the marker. The test class compiles a `@KsService` interface annotated with these and then asserts against the generated `RpcMethod` descriptor that:

- plain methods have empty metadata
- sibling annotation arguments (strings, ints, vararg-of-strings, booleans) are captured with the right types and values
- arguments left at their default values are NOT captured
- multiple sibling annotations on the same method are each captured separately
- annotations without the `@KsMethodMetadata` marker are ignored
- looking up absent annotations returns null

Both JVM and linuxX64 targets exercise this test (both passed locally).

## Test plan

- [ ] `./gradlew :ksrpc-core:jvmTest :ksrpc-core:linuxX64Test`
- [ ] `./gradlew :ksrpc-test:jvmTest :ksrpc-test:linuxX64Test` (new `MethodMetadataPropagationTest` included)
- [ ] `./gradlew apiCheck` (BCV dumps for ksrpc-api and ksrpc-core refreshed)
- [ ] `./gradlew :ksrpc-compiler-plugin:test` (still green — plugin still works against older ksrpc-core artifacts via the soft-lookup fallback)

Closes #11


Monkopedia:
Reworked per review feedback. Three new commits on top of 771948e:

- b50333f — [Refactor ServiceClass to use ServiceMethod wrapper](https://github.com/Monkopedia/ksrpc/pull/24/commits/b50333f3cb2b741c99a08e646f845426c8e14e2a): replaces the parallel `methods: List<Pair<...>>` / `methodMetadata: Map<...>` structures with a single `methods: List<ServiceMethod>`, where each entry owns its own `@KsMethod` annotation and sibling metadata annotations. Removes the implicit cross-structure invariant. Updated the three call sites in `StubGeneration`, `KsrpcIrGenerationExtension`, and the visitor that populated them.
- 460a2da — [Remove lookupKClass — drop runtime KClass reflection from KClassValue](https://github.com/Monkopedia/ksrpc/pull/24/commits/460a2da997f17d3cc44f5a211192899606c31a70): deletes the `expect fun lookupKClass` + JVM/native/JS/wasmJs actuals and the `KClassValue.kClass: KClass<*>?` getter. `classFqName: String` stays as the single field. No tests asserted on `.kClass`, so nothing to migrate there. The compiler plugin already constructed `KClassValue` with just the FQ name, so codegen is unchanged.
- dab4c25 — [Require metadata arg on RpcMethod — drop @JvmOverloads and default](https://github.com/Monkopedia/ksrpc/pull/24/commits/dab4c25911ab0c71d21706346e7b50adc70731d1): drops `@JvmOverloads` and the `= emptyList()` default. Plugin codegen now unconditionally emits the five-arg constructor call (passing `emptyList()` when there is no metadata) whenever the classpath ksrpc-core exposes `MethodMetadata`. The existing `metadataSupported` fallback to the four-arg shape stays so the plugin still works against a pre-#11 ksrpc-core (this is what the plugin's JUnit tests link against — removing the fallback broke them with `ksrpc-core on the compile classpath is missing MethodMetadata support`). Updated the hand-built `RpcMethod(...)` call sites in `ksrpc-test` to pass `metadata = emptyList()`. BCV dumps regenerated.

Verified:
- `./gradlew :ksrpc-test:jvmTest :ksrpc-test:linuxX64Test` — pass
- `./gradlew :ksrpc-compiler-plugin:test` (via compiler included build) — pass
- `./gradlew apiCheck` — pass (after `apiDump`)

ktlint note: `MetadataIrBuilder.kt` has pre-existing `blank-line-between-when-conditions` / `function-signature` violations from the original #11 commit, and there are unrelated pre-existing violations in `ReadWritePacketChannel.kt` and `NativeTestUtil.kt` on main. None of the three rework commits introduced new ktlint issues; I left the pre-existing ones alone per "don't touch anything unrelated."

Monkopedia:
Pushed 62a332b: `MetadataValue.KClassValue` and `MetadataValue.EnumValue` now hold real `KClass<*>` and `Enum<*>` references emitted directly by the compiler plugin from the annotation's `IrClassReference` / `IrGetEnumValue`, instead of FQ-name strings. Verified on JVM and linuxX64 (`:ksrpc-core` + `:ksrpc-test` tests, `:compiler:ksrpc-compiler-plugin:test`, `apiDump` / `apiCheck`).

Monkopedia:
Pushed c14784b: converted all MetadataValue subclasses and MethodMetadata itself to `data class`. Kept custom `toString()` on KClassValue/EnumValue/ListValue/MethodMetadata where the default would be uglier; dropped it on the primitive wrappers and StringValue. BCV dumps updated (added componentN/copy entries).

PR #25, opened 2026-04-19
Add UnhandledMethodHandler for receiver-side fallback dispatch

### LINKED ISSUE #15: UnhandledMethodHandler interface for opt-in catch-all dispatch on services
## Motivation

Some protocols are open-ended: vendor extensions, runtime-registered capabilities, or experimental methods introduce RPC method names that aren't present in the Kotlin service interface at compile time. Today ksrpc errors on any method name that doesn't match a registered endpoint.

Applications need a clean way to opt in to catch-all dispatch — to log, proxy, forward to another service, or gracefully ignore unknown calls — without monkey-patching dispatch internals.

## Proposed design

```kotlin
interface UnhandledMethodHandler {
    suspend fun onUnhandled(method: String, params: JsonElement?): JsonElement?
}
```

A service opts in by also implementing this interface:

```kotlin
class MyClientImpl : LanguageClient, UnhandledMethodHandler {
    override suspend fun onUnhandled(method: String, params: JsonElement?): JsonElement? {
        log.info("Unhandled method: $method")
        return null
    }
}
```

When ksrpc routes an incoming call to a service and no registered endpoint matches, it checks whether the service also implements `UnhandledMethodHandler` and dispatches there. If not, current error behavior is preserved.

## Design notes

- **Single unified handler regardless of transport.** This is a ksrpc-core concept, not a jsonrpc-specific one. Notify-vs-request semantics are a transport concern; at this API level the handler just returns a result, and the transport may or may not care about it (e.g. jsonrpc discards the return value when the inbound call was a notification).
- **Lives on the service, not the connection.** Connections can multiplex multiple services; each service's catch-all policy may differ, so it belongs on the service.
- **Opt-in, never automatic.** A service that doesn't implement the interface keeps the current strict behavior.

## Open design decisions

- Does `onUnhandled` get called for methods the interface *does* declare but the implementation forgot to override (abstract), or only for method names with no registered endpoint at all? Latter is simpler and more intuitive.
- Should the handler's exceptions be reported as `KsrpcException` over the wire, same as regular handler failures? Yes, seems obvious — worth stating.

## Motivating use case

Blocks LSP client work in `Monkopedia/kodemirror`. LSP allows vendor-specific notifications (e.g. `$/metals/status`), runtime-registered capabilities adding arbitrary new methods, and experimental protocols. A client wants to log-and-drop the unknown ones rather than tear down the connection.

Monkopedia:
## Summary

- New opt-in `UnhandledMethodHandler<T>` interface in `ksrpc-core`. Services can implement it alongside their normal `RpcService` to catch calls targeting endpoints that `findEndpoint` does not recognize, instead of the current unconditional `RpcEndpointException`.
- `HostSerializedServiceImpl.call` now catches `RpcEndpointException` from `findEndpoint`, checks whether the underlying service implements `UnhandledMethodHandler<S>`, and if so routes the call to `onUnhandled(method, input)`. Services that don't implement it keep the existing error behavior.
- Handler exceptions flow through the same outer `try/catch` in `HostSerializedChannelImpl.call` as regular handler exceptions, so they reach callers as remote errors.

## Approach — payload type

Used the existing `com.monkopedia.ksrpc.channels.CallData<T>` envelope. This is the transport-agnostic wrapper core's dispatch path already passes around (`SerializedService<T>.call(endpoint, input: CallData<T>)`) — it already accommodates both serialized strings (JSON transports) and binary (`ByteReadChannel`) payloads. No new sealed type was needed, no `JsonElement`, no `Any?`. Every transport benefits automatically because routing lives in the core dispatch path.

Interface signature:

```kotlin
interface UnhandledMethodHandler<T> {
    suspend fun onUnhandled(method: String, input: CallData<T>): CallData<T>
}
```

## Test plan

Tests live in `ksrpc-test/src/commonTest/` so they run across all four `RpcFunctionalityTest` transports (SERIALIZE, PIPE, HTTP, WEBSOCKET):

- [x] `RpcUnhandledMethodHandlerNotImplementedTest` — service without handler: unknown method yields `Unknown endpoint: missing` error, preserving current behavior.
- [x] `RpcUnhandledMethodHandlerImplementedTest` — service with handler: known endpoints still dispatch; unknown endpoint invokes `onUnhandled` and its serialized return value reaches the caller.
- [x] `RpcUnhandledMethodHandlerThrowsTest` — exception from `onUnhandled` is reported to the caller as a remote error.
- [x] `./gradlew :ksrpc-core:jvmTest :ksrpc-test:jvmTest`
- [x] `./gradlew :ksrpc-core:linuxX64Test :ksrpc-test:linuxX64Test`
- [x] `./gradlew :ksrpc-core:apiDump` and `apiCheck` (api diff committed)
- [x] ktlint

Closes #15

### DIFF
--- a/ksrpc-core/api/ksrpc-core.api
+++ b/ksrpc-core/api/ksrpc-core.api
@@ -288,6 +288,10 @@ public final class com/monkopedia/ksrpc/Transformer$DefaultImpls {
 	public static fun unpackError (Lcom/monkopedia/ksrpc/Transformer;Lcom/monkopedia/ksrpc/channels/CallData;Lcom/monkopedia/ksrpc/channels/SerializedService;)V
 }
 
+public abstract interface class com/monkopedia/ksrpc/UnhandledMethodHandler {
+	public abstract fun onUnhandled (Ljava/lang/String;Lcom/monkopedia/ksrpc/channels/CallData;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+}
+
 public abstract class com/monkopedia/ksrpc/channels/CallData {
 	public static final field Companion Lcom/monkopedia/ksrpc/channels/CallData$Companion;
 	public abstract fun isBinary ()Z
--- a/ksrpc-core/api/ksrpc-core.klib.api
+++ b/ksrpc-core/api/ksrpc-core.klib.api
@@ -90,10 +90,6 @@ abstract interface <#A: kotlin/Any?> com.monkopedia.ksrpc/KsrpcEnvironment { // 
     }
 }
 
-abstract interface <#A: kotlin/Any?> com.monkopedia.ksrpc/UnhandledMethodHandler { // com.monkopedia.ksrpc/UnhandledMethodHandler|null[0]
-    abstract suspend fun onUnhandled(kotlin/String, com.monkopedia.ksrpc.channels/CallData<#A>): com.monkopedia.ksrpc.channels/CallData<#A> // com.monkopedia.ksrpc/UnhandledMethodHandler.onUnhandled|onUnhandled(kotlin.String;com.monkopedia.ksrpc.channels.CallData<1:0>){}[0]
-}
-
 abstract interface com.monkopedia.ksrpc.channels/CancellationSupport { // com.monkopedia.ksrpc.channels/CancellationSupport|null[0]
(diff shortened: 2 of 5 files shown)


Monkopedia:
Rebased onto current main (includes PR #24 annotation propagation infrastructure). No conflicts occurred during rebase. All verification passed: ksrpc-core jvmTest/linuxX64Test, ksrpc-test jvmTest/linuxX64Test, and apiCheck.

Monkopedia:
Moved the type parameter from `UnhandledMethodHandler<T>` to the `onUnhandled` method per review feedback (commit ee735d1). API dump updated, all tests pass.

Issue #10, opened 2026-04-19
Native coroutine cancellation for RPC calls (client → cancel signal; server → handler Job cancelled)


Monkopedia:
## Motivation

Today, if the coroutine calling `execute()` on a JSON-RPC stub is cancelled, the `CompletableDeferred` awaiting a response is abandoned and no signal is sent to the remote side. On long-lived connections this leaks server-side work and state — the server keeps processing a request whose result nobody will read. Coroutine cancellation is idiomatic in Kotlin; it should propagate through RPC boundaries naturally rather than silently dropping on the floor.

## Proposed behavior

**Client side:** when the coroutine invoking a stub method is cancelled while awaiting a response, ksrpc signals cancellation to the remote side with the original request's id, then lets the `CancellationException` propagate.

**Server side:** when a cancellation signal arrives for a currently-handled request id, the `Job` running that handler is cancelled via standard coroutine cancellation.

Both sides: purely Kotlin-native. Consumers write ordinary coroutine code; cancellation just works.

## Design space — wire representation of "cancellation"

JSON-RPC 2.0 does not define cancellation. Concrete protocols invent their own conventions; LSP uses the notification `$/cancelRequest` with `{ id }`. Other JSON-RPC-based protocols may use different method names or have no cancel semantics at all.

Options for the implementer to weigh:

1. **Hard-code a convention** (e.g. LSP's `$/cancelRequest`). Fast; hostile to non-LSP users.
2. **Transport-level configuration** on `asJsonRpcConnection(...)` specifying the notification method name used for cancel. Opt-in; per-connection.
3. **Core-level primitive** ("cancel pending request X") in ksrpc-core, with ksrpc-jsonrpc taking a pluggable convention that translates to a specific wire message. Cleanest separation of concerns; more up-front design.

My lean is (3) but the implementer should decide.

## Scope

- Mechanism in ksrpc-core for cancellation propagation across RPC boundaries
- ksrpc-jsonrpc: emit and handle the cancel signal on the configured convention
- Tests covering both directions: client-cancel → server-handler-cancelled; server-cancel-signal → client-stub-throws (or completes early)

## Motivating use case

Blocks LSP client work in `Monkopedia/kodemirror` — `$/cancelRequest` is baseline functionality for any real-world language-server client (user types, triggering a completion; types again before the first completion returns; first request must be cancellable).

Monkopedia:
[...]

PR #26, opened 2026-04-19
Expose current RPC call (id, method) via coroutine context

### LINKED ISSUE #16: Expose current RPC call (id, method) via CoroutineContext inside handlers
## Motivation

Server-side handlers sometimes need to know their own request id or method name — for example, to emit follow-up notifications that reference the originating request (progress updates, partial results, streaming side-channels). Today this information is not accessible from within the handler coroutine; it is known to the dispatcher but not propagated.

## Proposed design

```kotlin
object CurrentRpcCall : CoroutineContext.Key<CurrentRpcCallElement>

class CurrentRpcCallElement(
    val id: JsonPrimitive?,  // null for notifications
    val method: String
) : CoroutineContext.Element {
    override val key: CoroutineContext.Key<*> get() = CurrentRpcCall
}

suspend fun currentRpcCall(): CurrentRpcCallElement? = coroutineContext[CurrentRpcCall]
```

ksrpc installs this element automatically in the coroutine context of every handler invocation. Handlers query it where needed:

```kotlin
override suspend fun longRunning(params: Params): Result {
    val call = currentRpcCall() ?: error("Not in an RPC call")
    reportProgress(call.id, "started")
    ...
}
```

## Design notes

- **Non-goal: cancellation.** This is not a request cancellation mechanism — see the separate native cancellation issue (#10). Handlers that want to co-operate with cancellation should use standard coroutine cancellation (`isActive`, `ensureActive`, suspension points).
- **Not propagated across calls.** When a handler fires off a further RPC call (even to the same connection), the new call is a fresh invocation; it does not inherit this element, and the nested stub invocation gets a clean coroutine context (or one managed by the stub itself).
- **`id` is protocol-typed.** For JSON-RPC specifically, `JsonPrimitive?` (null for notifications, string or integer otherwise) matches the wire type. If this element is generalized to other transports, the type may need to widen — implementer's call on the exact shape.

## Open design decisions

- Should the element also expose a reference to the current `SerializedService`, the caller's connection, or any other dispatcher-level context? Minimal scope says no — keep to id + method — but the mechanism could carry more.
- Is `CurrentRpcCallElement` the right name? Alternatives: `RpcCallContext`, `CurrentCall`, etc.

## Motivating use case

Blocks LSP client work in `Monkopedia/kodemirror`. LSP's progress reporting protocol requires server→client notifications that reference tokens given in the original request's params; handlers need access to context to correlate progress.

Monkopedia:
## Summary

Introduces `CurrentRpcCallElement` — a coroutine-context element carrying the in-flight RPC call's identity — plus a `currentRpcCall()` suspend accessor. Every transport dispatcher now installs this on the handler coroutine so server-side handlers can correlate follow-up traffic (progress notifications, streaming side-channels, partial-result pushes) with the originating request without the dispatcher having to wire it manually.

- `CurrentRpcCallElement(id: RpcCallId?, method: String)` in `com.monkopedia.ksrpc.channels`; `suspend fun currentRpcCall(): CurrentRpcCallElement?`.
- `JsonRpcWriterBase.launchRequestHandler` installs the element with `JsonRpcCallId(id)` (null id for notifications) and the request method.
- `PacketChannelBase.executeReceive` installs the element with `PacketCallId(channelId, messageId)` and the packet endpoint, for inbound request packets only.
- `RpcMethod.callChannel` strips the element from the caller's context before routing outbound stub calls, so a handler that fires a nested RPC does NOT leak its own call identity into the nested invocation — the nested inbound dispatch installs a fresh element on the remote side.

## Approach

**Reusing `RpcCallId` from #10.** #10 shipped `RpcCallId` (marker interface) with `PacketCallId(channelId: String, messageId: String)` in ksrpc-packets and `JsonRpcCallId(JsonPrimitive)` in ksrpc-jsonrpc, to key cancellation tracking. This change exposes that abstraction via the coroutine context and does NOT redefine it.

**Sealing note.** The task and the user-direction comment asked for `sealed interface RpcCallId`. Attempting `sealed` fails to compile: `PacketCallId` and `JsonRpcCallId` live in sibling Gradle modules (`ksrpc-packets`, `ksrpc-jsonrpc`), and Kotlin prohibits extending a sealed hierarchy from a different module. Collapsing the subtypes into `ksrpc-core` would violate the existing module boundary (transport-specific identifiers in transport modules). Kept as a non-sealed `interface` and expanded the doc to explain why — consumers that need a transport-specific view should `is` / cast to the subtype, which matches the pattern #10 already established inside `CancellationSupport`.

**Why strip in `callChannel`, not in the dispatch entrypoint.** `withContext(x)` merges `x` onto the caller's context — it cannot remove keys the caller already carries. The only way to strip `CurrentRpcCallElement` for the stub chain is to rebase on `coroutineContext.minusKey(...)`. Done once in `callChannel` (the stub-outbound path), with a comment covering the subtlety.

## Test plan

- [x] `./gradlew :ksrpc-core:jvmTest :ksrpc-packets:jvmTest :ksrpc-jsonrpc:jvmTest :ksrpc-test:jvmTest`
- [x] `./gradlew :ksrpc-core:linuxX64Test :ksrpc-packets:linuxX64Test :ksrpc-jsonrpc:linuxX64Test :ksrpc-test:linuxX64Test`
- [x] `./gradlew apiDump && ./gradlew apiCheck` — `.api` and `.klib.api` updated
- [x] `./gradlew ktlintCheck` on affected modules (pre-existing violations in `PacketCancellationTest.kt`, `RpcSubserviceLeakTest.kt`, `ReadWritePacketChannel.kt` are untouched by this change)

New test file `CurrentRpcCallTest.kt` covers:
- packet-transport handler sees a `PacketCallId` and the method name
- jsonrpc-transport handler sees a `JsonRpcCallId`
- jsonrpc notification handler sees a null id, method still populated (drives the notification path via a Unit-return stub method)
- nested stub call from within a handler sees a different (fresh) id, confirming the strip in `callChannel`
- handler that never touches `currentRpcCall()` continues to round-trip normally

Closes #16


Monkopedia:
Rebased onto current main (3337aab) which includes PR #24 (annotation propagation). Resolved one conflict in `ksrpc-core/api/ksrpc-core.klib.api` where both PRs added new classes at the same location — kept both `CurrentRpcCallElement` and `MethodMetadata`. Also corrected the JVM API dump (`ksrpc-core.api`) filename entry from `CurrentRpcCallKt` to `CurrentRpcCallElementKt`. All tests pass (ksrpc-core, ksrpc-test, ksrpc-jsonrpc, ksrpc-packets); apiCheck passes.

Monkopedia:
Reworked to address the code review. Force-pushed a single squashed commit (`340775f`) that replaces the previous PR contents on `issue-16-current-rpc-call`.

## What changed from the previous iteration

The element is now installed at exactly one central chokepoint — `RpcMethod.call()`, the single site that actually invokes `method.invoke(service, ...)`. Transports no longer install `CurrentRpcCallElement` themselves; they pass the wire-level call id explicitly through the dispatch signatures, and `RpcMethod.call()` does the install just before delegating to the handler.

### New shape

- `CurrentRpcCallElement(method: RpcMethod<*,*,*>, id: RpcCallId?)` — `method` is non-nullable (always installed around a concrete dispatch), `id` is nullable (null for in-process calls, HTTP, and jsonrpc notifications). Handlers now read `currentRpcCall()?.method.endpoint`, `.metadata`, etc. — the full `RpcMethod` is exposed.
- `RpcMethod.call(channel, service, input, callId)` — `callId: RpcCallId?` added as a required parameter. No defaults.
- `SerializedService.call(endpoint, input, callId)` and `SerializedChannel.call(channelId, endpoint, data, callId)` — `callId: RpcCallId?` added as a required parameter on both overloads. No defaults.
- `JsonRpcChannel.execute(method, message, isNotify, id: JsonPrimitive?)` — `id` added so the writer can forward the request id through to the serialized channel.
- `RpcMethod.callChannel` restored to its pre-PR shape (only the `withContext(channel.context.minusKey(Job))` from PR #10). No more `CurrentRpcCallElement` stripping — unnecessary, because each destination's `RpcMethod.call()` installs its own fresh element.

### Transport changes

- `JsonRpcWriterBase.launchRequestHandler` — no `CurrentRpcCallElement` install. It forwards the incoming `message.id` to `channel.execute(method, params, isNotify, id)`.
- `JsonRpcServiceWrapper.execute` — builds `JsonRpcCallId(id)` (if id != null) and passes it as `callId` to the underlying `SerializedService.call(...)`.
- `PacketChannelBase.executeReceive` — no `CurrentRpcCallElement` install. Builds `PacketCallId(channelId, messageId)` and passes it as `callId` to `serviceChannel.call(...)`.
- HTTP server routing passes `callId = null` (HTTP has no wire-level correlation id).
- Client-side stub path (`RpcMethod.callChannel` → `channel.call(...)`) passes `callId = null` (the stub allocates its own wire id at the transport layer; that's separate from the handler-visible id).

### BCV

This is a pre-1.0 BCV break — required parameters added, no defaults. `.api` and `.klib.api` updated via `apiDump` (committed). `apiCheck` passes.

### Verification

- `./gradlew :ksrpc-test:jvmTest :ksrpc-test:linuxX64Test` — pass
- `./gradlew :ksrpc-jsonrpc:jvmTest :ksrpc-packets:jvmTest :ksrpc-introspection:jvmTest` — pass
- `./gradlew :compiler:ksrpc-compiler-plugin:test` — pass (the plugin only emits `callChannel`, which kept its existing 2-arg signature, so no plugin changes were needed)
- `./gradlew apiCheck` — pass
- ktlint on all touched files — pass (remaining violations in the repo are pre-existing in untouched files)

### Notes

- The `UnhandledMethodHandler` fallback path merged in #25 was rebased cleanly: `HostSerializedServiceImpl.call` now takes `callId` and forwards it to `rpcEndpoint.call(this, service, input, callId)`, with the unhandled-method fallback preserved untouched around the endpoint lookup.
- The `MultiChannelRaceTest.testNoMultiChannelFailureWhenResponseArrivesAfterClose` and `PacketCancellationTest.cancellationBeforeResponseDoesNotLeakPendingState` tests are flaky under local JDK 21 `linuxX64Test` — they passed on re-run and on the pre-rework base, so the flake is unrelated to these changes.

Co-Authored-By: Claude Opus 4 (1M context) <noreply@anthropic.com>

PR #34, opened 2026-04-19
Add SerializedChannel overload to serveWebsocket for parity with serve

### LINKED ISSUE #22: ksrpc-ktor-websocket-server: add SerializedChannel overload to serveWebsocket for parity with serve
Minor symmetry issue surfaced by the #4 audit.

HTTP serve has two overloads:

```
public static final fun serve (Routing; String; SerializedChannel; KsrpcEnvironment)
public static final fun serve (Routing; String; SerializedService; KsrpcEnvironment)
```

WebSocket serve has only one:

```
public static final fun serveWebsocket (Routing; String; SerializedService; KsrpcEnvironment)
```

Either:
- Add `SerializedChannel` overload for WS so both transports accept the same arg shapes, or
- Intentionally drop the `SerializedChannel` overload on HTTP and document the reason.

This is also partly covered by #19 (transport naming unification) but calling it out separately so it doesn't get lost if #19 goes a different direction.

Refs #4, #19

Monkopedia:
## Summary
- Adds a `serveWebsocket(basePath, channel: SerializedChannel<String>, env)` overload mirroring the existing `serve` pair in `HttpStream.kt`, so callers that already own a dispatch-side channel can plug it into the websocket transport without unpacking the root service.
- The new overload reuses the existing `WebsocketPacketChannel.connect` wiring via a small `ChannelBackedSerializedService` adapter that prefers `ChannelClient.defaultChannel()` (preserving close hooks and sub-service routing for the common `HostSerializedChannelImpl` / `Connection` case) and falls back to dispatching against `ChannelClient.DEFAULT` when the supplied channel does not implement `ChannelClient`.
- Refreshes `ksrpc-ktor-websocket-server.api` / `.klib.api` to expose the new symbol and adds a parallel type-reference test for the overload alongside the existing one.

## Test plan
- [x] `./gradlew :ksrpc-ktor-websocket-server:apiDump :ksrpc-ktor-websocket-server:apiCheck`
- [x] `./gradlew :ksrpc-ktor-websocket-server:ktlintCheck`
- [x] `./gradlew :ksrpc-ktor-websocket-server:jvmTest`
- [x] `./gradlew :ksrpc-test:jvmTest --tests "*Websocket*"` (existing websocket transport coverage still green)

Closes #22

Generated with Claude Code


Monkopedia:
Deferring — the adapter approach doesn't match any existing pattern in the codebase. Doing this properly requires refactoring PacketChannelBase to accept an injectable dispatch target, which is a medium-size change. Filed as a follow-up; not blocking 1.0.0.

PR #35, opened 2026-04-19
@KsNotification sibling annotation for explicit notification opt-in

### LINKED ISSUE #12: @KsNotification sibling annotation (explicit opt-in for notifications)
## Motivation

The current "Unit-return ⇒ JSON-RPC notification" heuristic has known gaps flagged in the README's TODO:

- Requests that return `null` cannot use `Unit` and must be forced into `JsonElement?` shape, which is noise at call sites.
- A future protocol request that happens to return nothing useful can't share the same Kotlin idiom as a genuine notification.
- There is no compile-time signal distinguishing "fire-and-forget" from "awaited but returns no data", which are meaningfully different protocol semantics.

Explicit opt-in removes the ambiguity and lets the compiler enforce correctness.

## Proposed design

Introduce `@KsNotification` as a sibling annotation to `@KsMethod`:

```kotlin
@KsService
interface MyService : RpcService {
    @KsMethod("/subscribe")
    @KsNotification
    suspend fun subscribe(req: SubscribeReq)

    @KsMethod("/shutdown")
    suspend fun shutdown(u: Unit): JsonElement?  // request that returns null — explicit, no heuristic
}
```

### Rules

- Compile-time **error** if `@KsNotification` is applied to a method whose return type is not `Unit`. (Notifications by definition have no reply; a return value on the Kotlin side would never be populated.)
- ksrpc-jsonrpc reads the annotation and sends notifications (no `id`, no response awaited).
- Transports without a notification/request distinction (e.g. plain HTTP) are free to ignore it.

## Infrastructure

## Migration policy — open question

Two options; implementer chooses:

**(a) Hard break at 1.0.** A Unit-return method with only `@KsMethod` is no longer treated as a notification — it is a request whose response happens to carry no data. Existing users must add `@KsNotification` where they meant notification.

**(b) One-release deprecation.** Unit-return + `@KsMethod` with no `@KsNotification` continues to be treated as a notification but emits a compiler warning instructing the user to add the annotation. Remove the heuristic in the next release.

No preference from this issue's motivating user; pre-1.0 breaking change is acceptable.

## Motivating use case

Blocks LSP client work in `Monkopedia/kodemirror`. LSP has 26 notifications and 67 requests, several of which return nullable types (e.g. `shutdown`, `textDocument/willSaveWaitUntil`). The current heuristic is ambiguous for these.

Monkopedia:
## Summary
- Adds `@KsNotification` annotation in `ksrpc-api` as a `@KsMethodMetadata`-annotated sibling of `@KsMethod`, marking methods as JSON-RPC notifications (fire-and-forget, no response)
- Replaces the implicit "Unit return = notification" heuristic in `JsonRpcSerializedChannel` with explicit `RpcMethod.isNotification` metadata check
- Adds compile-time validation in the compiler plugin: `@KsNotification` on a non-Unit-return method emits a compilation error

## Approach
The old heuristic was in `JsonRpcSerializedChannel.call(RpcMethod, ...)` line 41:
```kotlin
call(endpoint.endpoint, input, !endpoint.hasReturnType)
```
This treated every Unit-returning method as a notification. Replaced with:
```kotlin
call(endpoint.endpoint, input, endpoint.isNotification)
```
where `isNotification` checks for the `@KsNotification` metadata entry propagated by the existing annotation infrastructure from PR #24.

**Migration (hard break, option (a)):** Unit-return methods without `@KsNotification` are now normal requests. Users relying on the old heuristic must add `@KsNotification` where they intended notification behavior.

## Test plan
- [x] **Compile-time validation:** `KsNotificationValidationTest` - `@KsNotification` on Unit-return compiles; on non-Unit-return fails with clear error
- [x] **Runtime metadata propagation:** `MethodMetadataPropagationTest` - `@KsNotification` method has metadata entry; `isNotification` returns true; plain/Unit-without-annotation methods return false
- [x] **Transport behavior:** `JsonRpcSerializedChannelNotifyTest` - Unit-return without `@KsNotification` uses request path (migration); with `@KsNotification` uses notify path
- [x] **Integration:** `CurrentRpcCallTest.jsonRpcNotificationHandlerSeesNullId` - updated to use `@KsNotification`, verifies end-to-end notification flow
- [x] API dump updated (`apiDump` + `apiCheck` pass)
- [x] ktlint passes on changed modules

Closes #12

### DIFF
--- a/compiler/plugin/src/main/java/FqConstants.kt
+++ b/compiler/plugin/src/main/java/FqConstants.kt
@@ -66,6 +66,7 @@ object FqConstants {
     val KS_SERVICE = FqName("com.monkopedia.ksrpc.annotation.KsService")
     val KS_INTROSPECTABLE = FqName("com.monkopedia.ksrpc.annotation.KsIntrospectable")
     val KS_METHOD_METADATA = FqName("com.monkopedia.ksrpc.annotation.KsMethodMetadata")
+    val KS_NOTIFICATION = FqName("com.monkopedia.ksrpc.annotation.KsNotification")
 
     val METHOD_METADATA = ClassId(FQPKG, Name.identifier("MethodMetadata"))
     val METADATA_VALUE = ClassId(FQPKG, Name.identifier("MetadataValue"))
--- a/compiler/plugin/src/main/java/KsrpcIrGenerationExtension.kt
+++ b/compiler/plugin/src/main/java/KsrpcIrGenerationExtension.kt
@@ -59,6 +59,9 @@ class KsrpcIrGenerationExtension(private val report: MessageCollector) : IrGener
                 serviceMethod.function,
                 serviceMethod.ksMethodAnnotation,
                 names
+            ) && validateNotification(
+                cls.irClass,
+                serviceMethod
             ) && current
         }
     }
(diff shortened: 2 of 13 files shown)


Monkopedia:
Reworked per review feedback in b591ec5: removed `RpcMethod.isNotification` from ksrpc-core and moved the helper to a private top-level extension in `ksrpc-jsonrpc`. Tests updated to assert on metadata lookup directly. BCV dumps updated. All verification passed (core/test/jsonrpc tests, apiDump, apiCheck).

PR #36, opened 2026-04-19
Add KsrpcException base class and @KsErrorData annotation

### LINKED ISSUE #13: KsrpcException base class + @KsErrorData sibling annotation for typed error payloads
## Motivation

When a server-side handler fails, the client today receives an `IllegalStateException` with a string message. Any structured error payload the server sent in JSON-RPC's `error.data` field is effectively lost — callers have no typed access.

Many JSON-RPC-based protocols define typed error data for specific methods. LSP, for instance, declares typed error payloads on `initialize` (`InitializeError` with a `retry` boolean), `textDocument/diagnostic`, and `workspace/diagnostic`. Clients need to be able to inspect these structurally.

### Base exception in ksrpc-core

```kotlin
open class KsrpcException(
    val code: Int,
    message: String,
    val data: JsonElement? = null
) : RuntimeException(message)
```

This becomes the type clients always catch when a remote call errors — replacing the current bare `IllegalStateException` / `RpcFailure` paths.

### Sibling annotation for typed payload

```kotlin
@KsMethod("/initialize")
@KsErrorData(InitializeError::class)
suspend fun initialize(params: InitializeParams): InitializeResult
```

When `@KsErrorData` is present, the stub decodes the incoming `error.data` using the declared class's serializer and gives callers typed access to it. Callers throwing can construct with the data already typed; ksrpc handles serialization.

### Throwing side

```kotlin
override suspend fun initialize(params: InitializeParams): InitializeResult {
    if (!compatible(params.clientInfo)) {
        throw KsrpcException(
            code = INITIALIZE_ERROR_CODE,
            message = "Client incompatible",
            data = Json.encodeToJsonElement(InitializeError(retry = true))
        )
    }
    ...
}
```

## Infrastructure

## Open design decisions

- **Typed exception shape**: parameterized `KsrpcException<T>` (simpler stub codegen, `catch` needs type-check) vs. per-method generated subclass (cleaner `catch (e: InitializeException)`, more generated code). Implementer's call.
- **Serializer resolution**: rely on `@Serializable` companion `.serializer()` via reflection, require the class to be passed with a serializer, or use `kotlinx-serialization`'s `serializer<T>()` reified variant in the generated code.
- **Interaction with unknown `code` values**: if the wire `error.code` doesn't match the expected range for the method, do we still attempt to decode `data`, or fall back to plain `KsrpcException`?

## Motivating use case

Blocks LSP client work in `Monkopedia/kodemirror`. LSP specifies typed error payloads on three methods today and more over time.

Monkopedia:
## Summary
- Introduces `KsrpcException(code: Int, message: String, data: String?, cause: Throwable?)` as the standard typed exception for remote RPC errors in ksrpc-api
- Adds `@KsErrorData(errorType: KClass<*>)` annotation with `@KsMethodMetadata` for declaring typed error payloads on `@KsMethod` functions
- Adds `RpcMethod.errorDataType` extension property to read the declared error type from metadata
- Updates `RpcException` and `RpcEndpointException` to extend `KsrpcException`
- Updates JSON-RPC error path to throw `KsrpcException` with code, message, and serialized data

## Approach
The `data` field on `KsrpcException` is `String?` -- the serialized form of the error payload. This keeps ksrpc-core transport-agnostic (no `JsonElement` or `Any?` in core types). Transport layers that support typed error data (e.g. JSON-RPC with `@KsErrorData`) can decode the string into the declared type on the client side.

## Test plan
- [x] `KsrpcExceptionTest`: verifies code/message/data/cause fields, RuntimeException inheritance, and that RpcException/RpcEndpointException extend KsrpcException
- [x] `KsErrorDataMetadataTest`: verifies `@KsErrorData` metadata propagation via compiler plugin and `errorDataType` extension
- [x] `JsonRpcKsrpcExceptionTest`: verifies JSON-RPC error responses throw KsrpcException with correct code/message/data
- [x] Updated existing tests (`JsonRpcErrorEnvelopeTest`, `JsonRpcOutOfOrderResponseTest`, `RpcEndpointExceptionTest`, `JsonRpcTest`) to expect KsrpcException
- [x] All existing tests pass (jvmTest, linuxX64Test, compiler plugin tests)
- [x] apiDump and apiCheck pass
- [x] ktlint passes on changed modules

## Follow-ups needed
- Compiler plugin validation that `@KsErrorData`'s class arg is `@Serializable` (skipped per scope guidance)
- JSON-RPC transport integration to automatically decode `error.data` using the declared serializer when `@KsErrorData` is present on the method

Closes #13

### DIFF
--- a/ksrpc-api/api/ksrpc-api.api
+++ b/ksrpc-api/api/ksrpc-api.api
@@ -1,4 +1,13 @@
-public final class com/monkopedia/ksrpc/RpcException : java/lang/RuntimeException {
+public class com/monkopedia/ksrpc/KsrpcException : java/lang/RuntimeException {
+	public fun <init> (ILjava/lang/String;Ljava/lang/String;Ljava/lang/Throwable;)V
+	public synthetic fun <init> (ILjava/lang/String;Ljava/lang/String;Ljava/lang/Throwable;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public final fun getCode ()I
+	public final fun getData ()Ljava/lang/String;
+	public fun getMessage ()Ljava/lang/String;
+	public fun toString ()Ljava/lang/String;
+}
+
+public final class com/monkopedia/ksrpc/RpcException : com/monkopedia/ksrpc/KsrpcException {
 	public fun <init> (Ljava/lang/String;)V
 	public fun getMessage ()Ljava/lang/String;
 }
@@ -12,7 +21,7 @@ public final class com/monkopedia/ksrpc/RpcFailure {
 	public fun equals (Ljava/lang/Object;)Z
 	public final fun getStack ()Ljava/lang/String;
 	public fun hashCode ()I
-	public final fun toException ()Ljava/lang/RuntimeException;
+	public final fun toException ()Lcom/monkopedia/ksrpc/RpcException;
 	public fun toString ()Ljava/lang/String;
 }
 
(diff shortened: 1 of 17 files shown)


Monkopedia:
Closing — the design needs rethinking. A generic KsrpcException(code, message, data) doesn't solve the real problem. What's needed is a mechanism to bind exception types to error codes/names, with the compiler plugin generating serializer bindings for the round-trip. See comment on #13.

PR #37, opened 2026-04-19
Add @KsTimeout annotation for client-side call timeouts

### LINKED ISSUE #14: @KsTimeout sibling annotation — per-method timeout via withTimeout
## Motivation

Long-lived RPC stubs have no per-method timeout today. Callers can wrap in `withTimeout { ... }` manually at every call site, but (a) this is noise that belongs on the service contract, and (b) without native cancellation propagation (see related issue) the abandoned pending state leaks on the server side.

Promoting timeout to a first-class annotation on the method makes it discoverable, self-documenting, and uniformly applied.

## Proposed design

```kotlin
@KsMethod("/slow")
@KsTimeout(30_000)  // milliseconds
suspend fun slow(req: Req): Result
```

The generated stub wraps the remote call in `withTimeout(30_000) { ... }`. On timeout, the coroutine is cancelled, and — together with the native cancellation feature — the cancellation propagates to the server.

## Infrastructure

## Close dependency

Closely related to the native cancellation issue (#10). Timeout is best implemented **after or together with** native cancellation: without cancellation propagation, timing out a call leaks pending state on both sides (pending deferred on client, running handler on server). With cancellation in place, timeout gets end-to-end correctness for free.

This issue does not require cancellation to land first, but the feature is noticeably less useful without it, and I'd recommend sequencing.

## Semantics caveats

- **In-process calls may not observe the timeout.** If a stub happens to be backed by an in-process service (no real transport), the body completes synchronously and `withTimeout` has nothing to interrupt. Acceptable; document.
- **On transports without cancellation propagation** (e.g. plain HTTP POST), timeout still cancels the local coroutine, but the server continues to completion. Acceptable for HTTP — that's always been its story.

## Open design decisions

- Unit: should it be `Long` milliseconds, `Duration`, or both via overloaded annotations?
- Does the annotation accept `-1` / `0` as "no timeout" to allow overriding at subclass declaration sites, or is absence the only way to express "no timeout"?

## Motivating use case

Blocks LSP client work in `Monkopedia/kodemirror`. Upstream `@codemirror/lsp-client` sets per-request timeouts for robustness against unresponsive or broken servers.

Monkopedia:
## Summary
- Adds `@KsTimeout(millis: Long)` annotation in `ksrpc-api` as a `@KsMethodMetadata` sibling annotation for `@KsMethod` functions
- Adds `RpcMethod.timeoutMillis` extension property that reads the timeout from captured metadata
- Wraps the existing `callChannel` body with `withTimeout(millis)` when a positive timeout is configured
- Combined with native cancellation (#10), timeout cancellation propagates to the server on supporting transports

## Approach
The `@KsTimeout` annotation is captured by the existing annotation propagation infrastructure (#11/PR #24) as `MetadataValue.LongValue`. At runtime, `RpcMethod.callChannel()` reads `timeoutMillis` and conditionally wraps the call body in `kotlinx.coroutines.withTimeout`. Values of 0 or negative are treated as "no timeout."

## Test plan
- [x] Metadata propagation: `@KsTimeout(5000)` on a method produces `rpcMethod.timeoutMillis == 5000L`
- [x] Metadata absence: method without `@KsTimeout` has `timeoutMillis == null`
- [x] Runtime success: call completing within timeout succeeds normally
- [x] Runtime timeout: call exceeding timeout throws an exception on the client
- [x] Runtime no-annotation: call without `@KsTimeout` proceeds without any timeout
- [x] All existing tests pass (`ksrpc-core` JVM/linuxX64, `ksrpc-test` JVM/linuxX64, compiler plugin)
- [x] `apiDump` + `apiCheck` pass (BCV updated for new annotation and extension property)

Closes #14

### DIFF
--- a/ksrpc-api/api/ksrpc-api.api
+++ b/ksrpc-api/api/ksrpc-api.api
@@ -59,6 +59,11 @@ public abstract interface annotation class com/monkopedia/ksrpc/annotation/KsNot
 public abstract interface annotation class com/monkopedia/ksrpc/annotation/KsService : java/lang/annotation/Annotation {
 }
 
+<<<<<<< HEAD
 public abstract interface annotation class com/monkopedia/ksrpc/annotation/KsrpcInternal : java/lang/annotation/Annotation {
+=======
+public abstract interface annotation class com/monkopedia/ksrpc/annotation/KsTimeout : java/lang/annotation/Annotation {
+	public abstract fun millis ()J
+>>>>>>> 1248868 (Add @KsTimeout annotation for client-side call timeouts)
 }
 
--- a/ksrpc-api/api/ksrpc-api.api
+++ b/ksrpc-api/api/ksrpc-api.api
@@ -59,16 +59,12 @@ public abstract interface annotation class com/monkopedia/ksrpc/annotation/KsNot
 public abstract interface annotation class com/monkopedia/ksrpc/annotation/KsService : java/lang/annotation/Annotation {
 }
 
-<<<<<<< HEAD
-public abstract interface annotation class com/monkopedia/ksrpc/annotation/KsrpcInternal : java/lang/annotation/Annotation {
-=======
 public abstract interface annotation class com/monkopedia/ksrpc/annotation/KsTimeout : java/lang/annotation/Annotation {
 	public abstract fun millis ()J
-<<<<<<< HEAD
->>>>>>> 1248868 (Add @KsTimeout annotation for client-side call timeouts)
-=======
 	public abstract fun minutes ()J
 	public abstract fun seconds ()J
->>>>>>> f3199f6 (Support seconds/minutes params on @KsTimeout for natural duration expression)
+}
+
+public abstract interface annotation class com/monkopedia/ksrpc/annotation/KsrpcInternal : java/lang/annotation/Annotation {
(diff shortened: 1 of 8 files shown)


Monkopedia:
Reworked per review: commit f3199f6. `@KsTimeout` now accepts `millis`, `seconds`, and `minutes` params (all defaulting to 0). Runtime computes total via `millis + seconds * 1000 + minutes * 60_000`. Tests updated for `seconds`, `minutes`, and combined usage.

Monkopedia:
Added timeout-triggered cancellation integration tests in commit 658f5e8.

**Tests added** (`KsTimeoutCancellationTest`):
- `timeoutCancelsRemoteHandlerOverPipeTransport` — verifies `@KsTimeout` fires, cancels the server handler via cancel frame, and connection remains healthy afterward.
- `timeoutCancelsRemoteHandlerOverJsonRpcWithLspConvention` — same verification over jsonrpc with `JsonRpcCancellationConvention.Lsp`.

**Result**: Both tests pass on JVM and linuxX64. Timeout-triggered cancellation propagates end-to-end correctly — the server handler's Job is cancelled when `@KsTimeout` fires, confirming the integration between #10 and #14 works as expected.

Issue #41, opened 2026-04-21
Generic @KsService interfaces: type parameters on services with serializer plumbing


Monkopedia:
The compiler plugin currently doesn't support type parameters on \`@KsService\` interfaces. Adding this unblocks #2 (Flow) and #39 (compiler Flow detection), and is a reasonable standalone feature.

## What to implement

Allow \`@KsService\` on generic interfaces:

\`\`\`kotlin
@KsService
interface KsStream<T> : RpcService {
    @KsMethod("/next")
    suspend fun next(u: Unit): T

    @KsMethod("/push")
    suspend fun push(item: T)
}
\`\`\`

At each usage site (sub-service return/param, or top-level service), the concrete type is known. The compiler generates:

- A parameterized \`RpcObject\` that takes a \`KSerializer<T>\` per type parameter in its constructor
- The \`RpcMethod\` for \`next\` uses the supplied serializer for the output transform
- The \`RpcMethod\` for \`push\` uses it for the input transform
- \`createStub(channel: SerializedService<S>): KsStream<T>\` returns a typed stub

Usage (at the wiring site where the concrete type is known):

\`\`\`kotlin
val obj: RpcObject<KsStream<Update>> = KsStream\$Object(Update.serializer())
\`\`\`

## Scope

- Compiler plugin: accept generic \`@KsService\` interfaces. Currently method-level type params are explicitly rejected (\`KsrpcIrGenerationExtension.validateMethod\`) — class-level ones aren't tested, probably need scoping.
- Generated \`RpcObject\` / stub: thread type parameters through codegen. Constructor takes \`KSerializer<T>\` per type param used in any method signature.
- Serializer use: when a method uses \`T\` directly (as in \`next(): T\`), the generated method's transform uses the injected serializer. When a method uses a generic wrapper (\`List<T>\`), the compiler needs to compose serializers — \`ListSerializer(tSerializer)\`.
- Tests in the compiler plugin unit tests: simple generic service with \`T\` return, \`T\` param, \`List<T>\` param. Verify generated code compiles and round-trips.
- Integration test in ksrpc-test: define a generic service, use it with a concrete \`T\`, exercise over PIPE.

## Non-goals

- Method-level type parameters (suspend fun <U> foo(): U) — still rejected.
- Variance (\`out T\`, \`in T\`) — start with invariant.
- Subservices with generic type params on their own — possible but not required for this issue. If it falls out naturally, good; if not, follow-up.

## Why priority:high

[...]

PR #42, opened 2026-04-21
Generic @KsService support (incremental): accept generic interfaces, parameterize Stub

Monkopedia:
## Summary

Incremental progress toward #41. The compiler plugin no longer rejects class-level type parameters on `@KsService` interfaces, and the generated `Stub` class is now parameterized with matching type parameters and a `KSerializer<T>` constructor parameter per type parameter.

This is **not** the complete feature — the `RpcObject`-bearing class that users would instantiate at wiring sites (e.g. `KsStream\$Object(String.serializer())`) is **not** emitted in this PR, and stub method bodies do not yet thread the injected serializer through call-channel transforms. The scope proved substantially larger than could land in a single review-sized change; this PR lays the FIR groundwork and unblocks subsequent work. See "Deferred" below for details.

## What's handled

- **Validation**: class-level type parameters on `@KsService` are accepted. Method-level type parameters are still rejected (unchanged). `out T` / `in T` variance is rejected with a clear diagnostic — only invariant type parameters are supported.
- **Stub class shape**: the generated `Stub` nested class now carries type parameters matching the service interface. The primary constructor takes the channel plus one `KSerializer<T>` per type parameter. For non-generic services the shape is unchanged (single `channel` parameter).
- **Companion behaviour**: for generic services the companion-object `RpcObject` is suppressed. Kotlin objects cannot declare type parameters, so the previous `Companion : RpcObject<KsStream<T>>` shape was malformed. Non-generic services continue to get a companion `RpcObject` exactly as before.
- **Unit tests** covering: generic `T` return, `T` param, `T?` return, `List<T>` param, method-level type params rejection, variance rejection, and the shape of the generic `Stub`.

## What's deferred (still #41)

- **`Object` nested class**: a non-companion nested class `Object<T>` with constructor `(tSerializer: KSerializer<T>)` implementing `RpcObject<KsStream<T>>`. This is the class users would instantiate as `KsStream.Object(String.serializer())`. The FIR and IR body generation for this class is the largest chunk of remaining work. Until it lands, generic services can be declared and have a Stub class, but there is no way for a caller to produce an `RpcObject<KsStream<T>>` for hosting or sub-service wiring.
- **Stub method body codegen**: the IR-side body generation for generic stub methods does not yet pull the injected `KSerializer<T>` out of the stub's fields to use in transforms. As a result, calling methods on a generic stub whose signature references `T` will fail at runtime. This is called out in the test comments.
- **Serializer composition**: for `T?`, `List<T>`, `Map<K, V>` etc. the compiler plugin needs to emit composed serializer expressions (`tSerializer.nullable`, `ListSerializer(tSerializer)`, ...). Not implemented in this PR.
- **Integration round-trip test**: a generic `@KsService` round-tripping over PIPE.

Suggest leaving #41 open with label `needs-decision` on whether to continue in-PR or split into separate tasks for `Object` generation, stub body wiring, and serializer composition.

## Test plan

- [x] `./gradlew :compiler:ksrpc-compiler-plugin:test` — passes including six new `GenericServiceTest` cases
- [x] `./gradlew :ksrpc-core:jvmTest :ksrpc-core:linuxX64Test` — passes
- [x] `./gradlew :ksrpc-test:jvmTest :ksrpc-test:linuxX64Test` — passes
- [x] `./gradlew apiCheck` — no public-API changes
- [x] ktlint clean on all modified files (pre-existing warning in generated BuildConfig.kt unchanged)

Partial delivery of #41.

### DIFF
--- a/compiler/plugin/src/main/java/CompanionGeneration.kt
+++ b/compiler/plugin/src/main/java/CompanionGeneration.kt
@@ -19,42 +19,31 @@ package com.monkopedia.ksrpc.plugin
 
 import org.jetbrains.kotlin.GeneratedDeclarationKey
 import org.jetbrains.kotlin.backend.common.extensions.IrPluginContext
-import org.jetbrains.kotlin.backend.common.lower.DeclarationIrBuilder
 import org.jetbrains.kotlin.backend.common.lower.irThrow
-import org.jetbrains.kotlin.descriptors.DescriptorVisibilities
-import org.jetbrains.kotlin.descriptors.Modality
 import org.jetbrains.kotlin.ir.backend.js.utils.isDispatchReceiver
 import org.jetbrains.kotlin.ir.builders.IrBlockBodyBuilder
 import org.jetbrains.kotlin.ir.builders.IrBuilderWithScope
-import org.jetbrains.kotlin.ir.builders.declarations.addConstructor
-import org.jetbrains.kotlin.ir.builders.declarations.addField
-import org.jetbrains.kotlin.ir.builders.declarations.addValueParameter
-import org.jetbrains.kotlin.ir.builders.declarations.buildClass
-import org.jetbrains.kotlin.ir.builders.declarations.buildReceiverParameter
-import org.jetbrains.kotlin.ir.builders.irBlock
 import org.jetbrains.kotlin.ir.builders.irBlockBody
 import org.jetbrains.kotlin.ir.builders.irBranch
 import org.jetbrains.kotlin.ir.builders.irCall
 import org.jetbrains.kotlin.ir.builders.irCallConstructor
-import org.jetbrains.kotlin.ir.builders.irDelegatingConstructorCall
 import org.jetbrains.kotlin.ir.builders.irElseBranch
 import org.jetbrains.kotlin.ir.builders.irEquals
 import org.jetbrains.kotlin.ir.builders.irGet
-import org.jetbrains.kotlin.ir.builders.irGetField
 import org.jetbrains.kotlin.ir.builders.irGetObject
+import org.jetbrains.kotlin.ir.builders.irIfThen
+import org.jetbrains.kotlin.ir.builders.irInt
+import org.jetbrains.kotlin.ir.builders.irNotEquals
 import org.jetbrains.kotlin.ir.builders.irReturn
-import org.jetbrains.kotlin.ir.builders.irSetField
 import org.jetbrains.kotlin.ir.builders.irString
 import org.jetbrains.kotlin.ir.builders.irVararg
 import org.jetbrains.kotlin.ir.builders.irWhen
 import org.jetbrains.kotlin.ir.declarations.IrClass
 import org.jetbrains.kotlin.ir.declarations.IrConstructor
 import org.jetbrains.kotlin.ir.declarations.IrDeclaration
-import org.jetbrains.kotlin.ir.declarations.IrDeclarationOrigin
 import org.jetbrains.kotlin.ir.declarations.IrParameterKind
 import org.jetbrains.kotlin.ir.declarations.IrProperty
 import org.jetbrains.kotlin.ir.declarations.IrSimpleFunction
-import org.jetbrains.kotlin.ir.declarations.IrValueParameter
 import org.jetbrains.kotlin.ir.expressions.IrBlockBody
 import org.jetbrains.kotlin.ir.expressions.IrBody
 import org.jetbrains.kotlin.ir.expressions.IrClassReference
@@ -62,15 +51,13 @@ import org.jetbrains.kotlin.ir.expressions.IrConstructorCall
 import org.jetbrains.kotlin.ir.expressions.IrExpression
 import org.jetbrains.kotlin.ir.symbols.IrClassSymbol
 import org.jetbrains.kotlin.ir.symbols.UnsafeDuringIrConstructionAPI
-import org.jetbrains.kotlin.ir.types.IrSimpleType
+import org.jetbrains.kotlin.ir.types.classFqName
 import org.jetbrains.kotlin.ir.types.defaultType
 import org.jetbrains.kotlin.ir.types.typeWith
 import org.jetbrains.kotlin.ir.util.SYNTHETIC_OFFSET
 import org.jetbrains.kotlin.ir.util.constructors
-import org.jetbrains.kotlin.ir.util.createThisReceiverParameter
 import org.jetbrains.kotlin.ir.util.kotlinFqName
 import org.jetbrains.kotlin.name.Name
-import org.jetbrains.kotlin.util.capitalizeDecapitalize.capitalizeAsciiOnly
 
 class CompanionGeneration(
     context: IrPluginContext,
@@ -129,6 +116,19 @@ class CompanionGeneration(
                     +irReturn(irListOfStrings(endpoints))
                 }
             }
+        // For generic service companions, eagerly emit the `arity` getter body — the
+        // property overrides RpcObjectFactory.arity and must return a fixed integer.
+        if (k.isGeneric) {
+            declaration.declarations
+                .filterIsInstance<IrProperty>()
+                .firstOrNull { it.name == FqConstants.ARITY }
+                ?.let { property ->
+                    val arity = cls.irClass.typeParameters.size
+                    property.getter?.body = context.irBuilder(property.getter!!).irSynthBody {
+                        +irReturn(irInt(arity))
+                    }
+                }
+        }
         declaration.isCompanion = true
         return emptyList()
     }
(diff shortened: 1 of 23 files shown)


Monkopedia:
## Continuation: runtime wiring for generic services

Follow-up to b13c468, landing the runtime side that makes generic `@KsService` actually usable.

### What's now working

- `Service(serializer)` returns an `RpcObject<Service<T>>`. Non-generic services are unchanged (`Service` still IS the `RpcObject`). The generated companion is always emitted — for generic services it's a plain object exposing `operator fun <T, ...> invoke(serializers): RpcObject<Service<T, ...>>`.
- The companion's `invoke` body constructs a generated nested `Obj<T, ...>` class that implements `RpcObject<Service<T, ...>>`. `Obj` is an internal implementation detail — callers only see `Service(ser)`.
- Stub method bodies for generic services are wired to the injected serializer: the generated method constructs an `RpcMethod` inline using the stub's per-instance `KSerializer<T>` field and calls `callChannel`.
- Serializer composition: `T` and `T?` are supported. `T?` calls `kotlinx.serialization.builtins.KSerializer.nullable`.
- **Integration round-trip test**: `ksrpc-test/commonTest/GenericServiceTest.kt` defines `GenericEcho<T>` with `echo(T): T` and `maybe(T?): T?`, implements it for `String`, and verifies the round-trip over SERIALIZE, PIPE, HTTP, and WEBSOCKET via `RpcFunctionalityTest`.

### Two small IR notes

- Executors (the anonymous `ServiceExecutor` classes used by `RpcMethod`) are pre-created in `StubGeneration.generateChildrenForClass` and stashed on `ServiceClass.genericExecutors`. Emitting them inside `generateBodyForFunction` threw `ConcurrentModificationException` because the IR visitor iterates class declarations while body generation runs.
- Stub overrides on generic services substitute the service's class-level type parameters with the Stub's via `IrTypeSubstitutor`. Without this, Kotlin/Native reports `Abstract function 'echo' is not implemented in non-abstract class 'Stub'`; the JVM didn't mind because of erasure.

### Limited

- `List<T>` / `Set<T>` / `Map<K, V>` wrappers of a type parameter are not yet composed into nested serializers. The FIR validation accepts `List<T>` (the compile test `generic ksservice with List of T compiles` passes) but the IR for such a method falls back to `serializer<T>()` on `List<T>`, which will fail at runtime. Follow-up work to emit `ListSerializer(tSer)` / `SetSerializer(tSer)` / `MapSerializer(kSer, vSer)` at IR time is straightforward given the plumbing already in `GenericMethodIrBuilder.buildSerializer`.
- A generic service used as a sub-service output type (e.g. `suspend fun child(): KsStream<T>`) isn't supported — the subservice transformer still assumes a companion-object `RpcObject`.

Both limitations are tracked on #41.

### Verification

- `./gradlew :compiler:ksrpc-compiler-plugin:test` — existing 6 `GenericServiceTest` cases plus everything else pass
- `./gradlew :ksrpc-test:jvmTest :ksrpc-test:linuxX64Test` — pass, including the new `GenericServiceRoundTripTest`
- `./gradlew apiCheck` — no API changes

Monkopedia:
Added `RpcObjectFactory<T : RpcService>` in ksrpc-core and wired the generic-companion codegen so it now implements `RpcObjectFactory<Service<*, *, ...>>`.

- New `RpcObjectFactory` interface: `arity: Int` + `create(List<KType>): RpcObject<T>`.
- New `resolveSerializerOrThrow(KType, serviceName)` helper (`@KsrpcInternal`) translates `kotlinx.serialization` failures into `IllegalArgumentException("Type argument $type used with $serviceName must be @Serializable", cause)`.
- FIR: generic companion declares `RpcObjectFactory<Service<*, *, ...>>` as supertype and exposes `arity` (override val) and `create` (override fun). Generation is gated on `RpcObjectFactory` being on the compile classpath, so plugin tests that pin ksrpc-core 0.9.8 keep passing.
- IR: eager `arity` getter body emits the fixed type-parameter count. `create` validates `typeArgs.size == arity` (throws `IllegalArgumentException` naming the service) and constructs the generated `Obj<Any?, Any?, ...>` with the resolved serializers.
- Non-generic services unchanged: the companion remains an `RpcObject<Service>` directly.
- BCV dumps (both `.api` and `.klib.api`) updated and committed.

Tests added in `ksrpc-test/src/commonTest/kotlin/GenericServiceTest.kt` (`GenericServiceFactoryTest`):
- `arity` matches declared type-parameter count.
- Companion cast to `RpcObjectFactory<*>` round-trips service name + endpoints (registry use case).
- `GenericEcho.create(listOf(typeOf<String>()))` round-trips a call through PIPE.
- Wrong arity (too few / too many) throws `IllegalArgumentException` naming the service.
- Non-serializable type argument (`data class NotSerializable` with no `@Serializable`) throws `IllegalArgumentException` mentioning both the type and the service.

Verification run:
- `:compiler:ksrpc-compiler-plugin:test`
- `:ksrpc-core:jvmTest` and `:ksrpc-core:linuxX64Test`
- `:ksrpc-test:jvmTest` and `:ksrpc-test:linuxX64Test`
- `apiDump` / `apiCheck`
- ktlint on all touched files

No platform-specific gotchas so far — JVM and linuxX64 both exercise the `kotlinx.serialization.serializer(KType)` path via the tests, and the non-serializable path correctly raises `IllegalArgumentException`. If wasmJs/JS become a concern (no kotlin-reflect), the helper already catches both `SerializationException` and `IllegalArgumentException` from the underlying resolver.

Monkopedia:
Follow-up: reified helpers now work for generic services.

- `@RpcObjectKey` loosened to `KClass<*>` so it can reference either an `RpcObject` (non-generic companion) or an `RpcObjectFactory` (generic companion); all four platform `actual` declarations updated.
- Compiler plugin now emits `@RpcObjectKey` for generic services too, pointing at the factory companion (the previous skip-for-generic branch is gone).
- `rpcObject<T>()` on all platforms (JVM, Native, JS, wasmJs) now inspects the resolved companion: returns it directly if it's an `RpcObject`, otherwise resolves `typeOf<T>()` arguments and delegates to `RpcObjectFactory.create(typeArgs)`.
- `serialized(env)` and `toStub()` already funneled through `rpcObject<T>()`, so they work for generic services with no further changes.
- Tests: `GenericServiceRoundTripTest` now exercises the reified path end-to-end via `impl.serialized<Generic<Arg>, _>(env)` / `channel.toStub<Generic<Arg>, _>()`, with dedicated assertions for `rpcObject<GenericEcho<String>>()` and the full serialized round-trip. The factory-focused tests (arity/serializer validation) stay to cover `create(...)` directly.
- BCV: klib dump updated for the `@RpcObjectKey` parameter type change.

Verified: `:compiler:ksrpc-compiler-plugin:test`, `:ksrpc-core:jvmTest`, `:ksrpc-core:linuxX64Test`, `:ksrpc-test:jvmTest`, `:ksrpc-test:linuxX64Test`, `apiCheck` all green.

Monkopedia:
Pushed fixes for the three reviewer items:

**Fix 1 — Cleanup fully-qualified `ConeKotlinType`**
Added imports in `FirKsrpcObjGenerator.kt` and `FirCompanionDeclarationGenerator.kt` and replaced the two spelled-out lambda return types with the short name. No behavior change.

**Fix 2 — JVM `rpcObject()` supertype walking**
When `T::class` has no directly-usable companion, the JVM actual now walks `T`'s supertypes via `kotlin.reflect.KClass.allSupertypes` (which preserves concrete type arguments — e.g. `GenericEcho<String>` on `TypedStream : GenericEcho<String>`). For each supertype we check the classifier's companion: an `RpcObject` is returned directly; an `RpcObjectFactory` is invoked with the type arguments from THAT supertype's `KType`, not from `typeOf<T>().arguments`.

**Fixes 3 & 4 — Tests**
- `reifiedRpcObjectRoundTrip` (commonTest): explicit end-to-end round trip through `rpcObject<GenericEcho<String>>()`, covering the full send/receive path beyond the metadata check in `reifiedRpcObjectResolvesFactory`.
- `rpcObjectResolvesGenericSupertype` (jvmTest): exercises the supertype-walking path with `interface TypedGenericEchoJvm : GenericEcho<String>` and does a full round-trip over a serialized channel.

**Two limitations surfaced while adding Fix 3 — documented in comments and follow-ups:**

1. **`@KsService` on a subtype-of-generic-parent crashes the plugin.** Marking `TypedGenericEcho : GenericEcho<String>` with `@KsService` triggers `Invalid synthetic declaration for ...TypedGenericEcho` from `CompanionGeneration`. The FIR generator happily creates a companion, but the IR-side `ServiceClass.findServices` + `validateClass` only treats classes with a *direct* `RpcService`/`IntrospectableRpcService` super as valid services, so the subtype is dropped from the map before `CompanionGeneration` runs. Fixing this requires either widening the validation to walk up to a `@KsService` ancestor or teaching the companion/stub generator to delegate endpoint/method generation to the parent. Tracked as a follow-up. For now, the subtype is kept as a plain Kotlin interface in the test — the supported pattern is "parent has `@KsService`, subtype is a normal inheriting interface, resolve through `rpcObject()` supertype walking".
2. **Subtype-specialization through `rpcObject<Subtype>()` is JVM-only.** Native/JS/WASM use `findAssociatedObject<RpcObjectKey>()`, which does not walk supertypes; without a compiler-plugin change to emit `@RpcObjectKey` on subtypes or a runtime walk, this path is unavailable there. Documented in the JVM actual's KDoc and in the JVM-only test file. Callers on those platforms should use `rpcObject<ParentGeneric<T>>()` directly.

**Verification**
- `:compiler:ksrpc-compiler-plugin:test` — pass
- `:ksrpc-core:jvmTest :ksrpc-test:jvmTest` — pass (including the two new tests)
- `:ksrpc-core:linuxX64Test :ksrpc-test:linuxX64Test` — pass
- `apiDump` produced a tiny addition (`allSupertypesPreservingArguments` helper, `@PublishedApi internal`) — committed.
- `apiCheck` — pass
- ktlint — no new violations from these changes (several pre-existing on the branch in unrelated files).

PR #89, opened 2026-04-23
Fix #19: unify transport entry-point naming for 1.0.0

### LINKED ISSUE #19: Unify transport entry-point naming: client/server asConnection vs asWebsocketConnection vs asJsonRpcConnection, serve vs serveWebsocket
From #4 audit. Each transport has a slightly different entry point name and shape, and they'll all freeze at 1.0.0.

## Current state

Client:
- HTTP (`ksrpc-ktor-client`): `HttpClient.asConnection(path, env): ChannelClient`  - **non-suspend**, returns **`ChannelClient`**, not `Connection`
- WebSocket (`ksrpc-ktor-websocket-client`): `HttpClient.asWebsocketConnection(path, env): Connection` (suspend)
- JsonRPC (`ksrpc-jsonrpc`): `Pair.asJsonRpcConnection(env, boolean): Connection` (suspend), `ProcessBuilder.asJsonRpcConnection(env)`, `stdInJsonRpcConnection(env)`
- Sockets (`ksrpc-sockets`): `Pair<ByteReadChannel, ByteWriteChannel>.asConnection(env): Connection` (suspend), plus a non-suspend `Pair.asConnection(scope, env): Connection`, plus `ProcessBuilder.asConnection(env)`, plus `withStdInOut(env, block)`.

Server:
- HTTP: `Routing.serve(path, SerializedChannel, env)` and `Routing.serve(path, SerializedService, env)`
- WebSocket: `Routing.serveWebsocket(path, SerializedService, env)` (no `SerializedChannel` overload)

## Problems

1. **Asymmetric suffixes**: WS and JsonRPC carry the transport suffix (`asWebsocketConnection`, `asJsonRpcConnection`) but HTTP and sockets both use bare `asConnection`. Either all of them should carry their suffix, or none should. Overload-by-receiver is what's disambiguating today, and that only works because the receiver types happen to be distinct.
2. **HTTP client returns `ChannelClient`, everything else returns `Connection`**. That's a real behavioural difference (HTTP is stateless - no server-initiated calls, no host side), but the return-type split makes it non-obvious to users. At minimum this should be documented in the KDoc; better would be a distinct name (`asHttpChannelClient` or similar) to make it clear that HTTP is not a full bidirectional connection.
3. **HTTP client `asConnection` is the only non-suspend variant.** Everyone else suspends. That's because HTTP doesn't have a handshake, but again it's silent in the signature.
4. **Server side is asymmetric**: HTTP `serve` accepts both `SerializedChannel` and `SerializedService`; `serveWebsocket` only accepts `SerializedService`. Either add the `SerializedChannel` overload to WS, or drop it from HTTP, or document why they differ.

## Suggestion

Pick one convention and apply it everywhere:

- **Option A**: All transports use `asXxxConnection` with their suffix; HTTP client becomes `asHttpChannelClient`; servers become `serveHttp` / `serveWebsocket` / `serveJsonRpc`.
- **Option B**: Everything uses bare `asConnection`/`serve`, disambiguated by receiver type; HTTP client uses a distinct name because its return type differs.

Either is fine. Option A is what most users will grep for. Option B is shorter. Either way, decide and align the whole surface before freeze.

Refs #4

Monkopedia:
## Summary

- Renames every client and server transport entry point to carry its transport suffix so the convention is consistent across HTTP, WebSocket, JSON-RPC, and sockets.
- Hard rename (no deprecated shims) since we are still pre-1.0.

## Renames (old -> new)

Client side:
- `HttpClient.asConnection(url, env)` -> `HttpClient.asHttpChannelClient(url, env)` (still returns `ChannelClient`, now named to reflect that)
- `Pair<ByteReadChannel, ByteWriteChannel>.asConnection(env)` -> `.asSocketConnection(env)` (both the suspend and the scope-taking overloads)
- `Pair<InputStream, OutputStream>.asConnection(env)` -> `.asSocketConnection(env)` (JVM)
- `ProcessBuilder.asConnection(env)` -> `.asSocketConnection(env)` (JVM, in `ksrpc-sockets`)

Server side:
- `Routing.serve(basePath, ...)` -> `Routing.serveHttp(basePath, ...)` (all three overloads: inline reified, `SerializedService`, `SerializedChannel`)

Unchanged (already consistent):
- `HttpClient.asWebsocketConnection(url, env)`
- `Pair<ByteReadChannel, ByteWriteChannel>.asJsonRpcConnection(env, ...)` and its JVM `InputStream`/`OutputStream` and `ProcessBuilder` siblings, and `stdInJsonRpcConnection(env)`
- `Routing.serveWebsocket(basePath, ...)`
- `JniConnection` (class-based, already transport-prefixed)
- `withStdInOut(env, block)` (lives in `ksrpc-sockets`; the stdin/stdout helper name is already transport-neutral)

## Deprecated shims

None. Pre-1.0, hard rename is cleaner.

## Call sites updated

- `ksrpc-bench`: 3 benchmark files.
- `ksrpc-server`: `BaseServiceApp`, `ServiceApp`.
- `ksrpc-test`: 9 commonTest files, 2 jvmTest files, 1 nativeTest util.
- `ksrpc-ktor` server test: `KtorServerModuleTypeReferenceTest`.
- `README.md`: sample code and the linked API surface list.

## BCV

`apiDump` run; `apiCheck` passes. Diffs limited to the three touched modules (`ksrpc-ktor-client`, `ksrpc-ktor-server`, `ksrpc-sockets`).

## Test plan

- [x] `./gradlew :ksrpc-core:jvmTest :ksrpc-core:linuxX64Test`
- [x] `./gradlew :ksrpc-test:jvmTest :ksrpc-test:linuxX64Test`
- [x] `./gradlew :ksrpc-jsonrpc:jvmTest :ksrpc-jsonrpc:linuxX64Test`
- [x] `./gradlew :ksrpc-ktor-client:jvmTest`
- [x] `./gradlew :compiler:ksrpc-compiler-plugin:test`
- [x] `./gradlew apiDump && apiCheck`
- [x] ktlint on touched files (no new violations; pre-existing unrelated violations left alone)

## Surprises / notes

- `JniConnection` in `ksrpc-jni` is a class with constructors, not an extension entry point, so its naming was already transport-prefixed and is untouched.
- `stdInConnection` and `withStdInOut` are not transport-specific names (they describe the stdin/stdout use case), so they were left alone.
- The jsonrpc `ProcessBuilder`/`InputStream`/`OutputStream`/`stdIn` helpers were already on the `asJsonRpcConnection` / `stdInJsonRpcConnection` naming, so no work there.
- Pre-existing unrelated ktlint failures remain in `ksrpc-api/.../KsMethod.kt`, `ksrpc-core` js/native source sets, and a handful of test files; they are not touched by this PR.

Closes #19

Generated with [Claude Code](https://claude.com/claude-code)


Monkopedia:
Reworked per review feedback (commit `07f18fe`).

Reverted the sockets `asSocketConnection` renames back to `asConnection` since in each case the receiver type itself is already the socket/stream transport, so the prefix was redundant:

- `Pair<ByteReadChannel, ByteWriteChannel>.asConnection`
- `Pair<InputStream, OutputStream>.asConnection` (JVM)
- `ProcessBuilder.asConnection`

Kept:
- `HttpClient.asHttpChannelClient(url, env)` — receiver is multi-protocol; symmetric with `asWebsocketConnection` on the same receiver.
- `Routing.serveHttp(...)` — generic ktor receiver; prefix disambiguates vs `serveWebsocket`.

Verification: `apiDump` regenerated (sockets API now shows `asConnection` again), `:ksrpc-sockets:jvmTest`, `:ksrpc-sockets:linuxX64Test`, `:ksrpc-test:jvmTest`, `:ksrpc-test:linuxX64Test`, `:compiler:ksrpc-compiler-plugin:test`, and `apiCheck` all pass. ktlint on touched modules has only pre-existing failures unrelated to these changes.

Issue #28, opened 2026-04-19
Per-call auth token: client-side provider, server-side context exposure


Monkopedia:
Add a first-class auth hook: a single client-side handler provides a token on each outbound call, and the token is available on the server side via the coroutine context (naturally alongside \`CurrentRpcCallElement\`).

## Motivation

ksrpc has no built-in auth story today. Consumers have to bolt auth on at the transport layer (ktor interceptors) which doesn't carry through packet-based or stdin/out transports, and doesn't give handlers a clean transport-agnostic place to read the credential. At 1.0.0 this is a noticeable gap for anything beyond in-process / trusted-network use.

## Proposed shape

**Client side — a provider configured per connection:**

```kotlin
val connection = HttpClient { }.asConnection(url, env) {
    authTokenProvider = { /* suspend function returning current token */ }
}
```

The provider runs before every outbound call. Return \`null\` → no token on that call. Should be allowed to suspend (for refresh flows).

**Server side — available via coroutine context:**

```kotlin
override suspend fun getSecret(u: Unit): Secret {
    val call = currentRpcCall() ?: error(...)
    val token = call.authToken ?: throw KsrpcException(..., \"missing auth\")
    verifyToken(token)
    return ...
}
```

\`authToken\` sits on \`CurrentRpcCallElement\` alongside the call id and the \`RpcMethod\` reference — fits the context-element design direction emerging from PR #26 and the feedback to fold the full \`RpcMethod\` into that element.

## Interaction with in-flight work

- **PR #26** (\`CurrentRpcCallElement\`) and the follow-up feedback to fold the full \`RpcMethod\` into it: \`authToken\` becomes another field on the same element. Design this together with #16's rework rather than bolt on after.
- **#13** (\`KsrpcException\`): a standard auth-failure exception code would be nice-to-have. Not blocking.
- **#17 / #18** (internal/experimental opt-in annotations): anything half-baked here can be gated behind \`@ExperimentalKsrpc\` until we're sure.

## Open design questions (\`needs-decision\` scope)

- **Wire representation per transport:**
  - HTTP: \`Authorization\` header is obvious.
  - Packet-based (sockets / ktor ws / stdin-out via ksrpc-packets): new packet field vs sideband packet vs reserved metadata slot?
  - JSON-RPC: reserved param? A \`meta\` object on the request? Protocol-specific convention (LSP has no auth, so this stays pluggable)?
[...]

PR #98, opened 2026-04-23
Fix #70: extensible packet format with TLV metadata section

### LINKED ISSUE #70: Extensible packet format: TLV metadata section for forward-compat before 1.0.0 freeze
## Problem

The current ksrpc packet format has fixed fields only. Every time we've needed to add a new field, it's been a wire break:

- #10 added the cancel frame (new bit in the type bitfield).
- #28 (context propagation) needs a metadata map on each request packet.
- Potential future additions (tracing, sub-protocol hints, etc.) would each break the wire again.

Once 1.0.0 freezes the format, every such break becomes a 2.0 event or requires version negotiation. Painful.

## Proposed fix

Add a **TLV metadata section** to the packet — type-length-value encoded fields that unknown peers can skip by length.

### Structure (approximate)

\`\`\`
[existing core fields]    # unchanged: type bitfield, channel/message id, endpoint, data
[metadata count: uint8]   # how many TLV fields follow (0 = no metadata)
[TLV 1: tag: uint16, length: uint16, value: bytes]
[TLV 2: ...]
...
\`\`\`

Parsing:
- Old peer reading new packet: reads core fields, sees \`metadata count\`, skips each TLV by its length. No crash.
- New peer reading old packet: same code path works (old peer writes \`metadata count = 0\`); new fields are simply absent.

### Forward compat guarantees established

- Adding a new TLV tag (auth, trace, N more) post-1.0.0 does NOT break existing peers.
- Removing a tag is still a break (old peers may expect it).
- The \`metadata count\` and TLV structure is the part that's frozen at 1.0.0.

## Integrates with

- **#28 (context propagation):** the wire-level "metadata map" on packet requests becomes a set of TLVs. Each context binding's \`wireKey\` maps to a reserved tag number (or stringly-tagged — see open question below).
- **#10 (cancellation):** the cancel frame is already bit-based in the type bitfield, so already forward-compatible. No change needed.

## Open questions

1. **Tag space:** numeric tag IDs (compact, registry-managed) or string tags (open-ended, bigger)? Numeric is cheaper on the wire. But reserving a numbered registry means the library decides tag assignments, which is a tighter coupling than stringly.
2. **Metadata on response packets too** or only requests? Responses would let bidirectional context (#28's deferred bit) land cleanly.
3. **Known-tag validation:** should unknown tags be silently ignored, or opt-in strict-mode that logs them?

## Why priority:high

This must land BEFORE 1.0.0 freezes the packet format — after that, future field additions become wire breaks. Blocking for 1.0.0.

## Implementation scope

- Extend \`Packet\` data class in ksrpc-packets to carry the metadata map.
- Serialization/deserialization: update \`Packet\` serializer to emit the count + TLV list, handle unknown tags on read.
- Tests: explicitly verify forward-compat by constructing a "new" packet with an extra TLV and decoding it with the existing parser — assert no crash, unknown data ignored.
- \`PacketChannelBase\` exposure of the metadata to higher layers (for #28's consumption).

Monkopedia:
## Summary

- Adds a `metadata: Map<Int, ByteArray>` section to `Packet` so future fields (auth, tracing, context propagation from #28, etc.) can be added without peers-must-upgrade-together wire breaks.
- Empty metadata is omitted from the wire via `@EncodeDefault(Mode.NEVER)`, so packets without metadata are byte-identical to the pre-TLV format.
- Introduces `PacketTlv` as a reserved tag registry; `CONTEXT_MAP = 1` is pre-reserved for #28. Tags are 16-bit; unknown tags are silently preserved/ignored.

## One-time wire break

The packet format changes in this PR between pre-1.0 and this release — a new serialised key (`x`) is introduced. An old peer receiving a new packet that contains a non-empty metadata map will reject it unless configured with `ignoreUnknownKeys`. Because empty metadata is not encoded, pre-existing deployments that don't opt into any metadata continue to interoperate unchanged.

**From this PR forward**, new metadata TLVs can be added without breaks: new peers write new tags, old peers that already understand the `x` key simply skip the ones they don't know. This is the forward-compat guarantee #70 exists to establish before the 1.0.0 format freeze.

## Implementation notes

- The TLV concept maps to an extensible `Map<Int, ByteArray>` field because `Packet` is encoded via the user-supplied `CallDataSerializer` (JSON by default, but pluggable). Rather than imposing a length-prefixed binary layout — which would fight every non-binary format — the TLV section rides on the chosen serializer's native map encoding. Forward-compat is preserved because consumers look up tags they know about and leave unknown entries alone.
- `PacketChannelBase` doesn't need a change: the new field defaults to empty, so existing send/receive paths already carry it correctly.
- No production consumers set metadata in this PR — #28 will fill it.

## TLV shape chosen

- Tag: 16-bit (`Int`, validated by convention via `PacketTlv.MAX_TAG = 0xFFFF`).
- Length: implicit (delegated to the underlying serializer's ByteArray encoding).
- Value: raw bytes, opaque.
- Count: implicit (serializer-native map size; JSON writes the full map).

## Test plan

- [x] `./gradlew :ksrpc-packets:compileKotlinJvm :ksrpc-packets:compileKotlinLinuxX64`
- [x] `./gradlew :ksrpc-test:jvmTest :ksrpc-core:jvmTest :compiler:ksrpc-compiler-plugin:test`
- [x] `./gradlew :ksrpc-test:linuxX64Test`
- [x] `./gradlew apiDump && apiCheck`
- [x] ktlint clean on touched files (`PacketMetadataTlvTest.kt`, `Packet.kt`, `PacketTlv.kt`).
- [x] New `PacketMetadataTlvTest` covers: empty-metadata omission, round-trip preservation, forward-compat with an unknown tag (99), pre-TLV wire decode yielding empty metadata.

Closes #70

Generated with Claude Code

### DIFF
--- a/ksrpc-packets/api/ksrpc-packets.api
+++ b/ksrpc-packets/api/ksrpc-packets.api
@@ -5,17 +5,18 @@ public final class com/monkopedia/ksrpc/packets/internal/ConstantsKt {
 
 public final class com/monkopedia/ksrpc/packets/internal/Packet {
 	public static final field Companion Lcom/monkopedia/ksrpc/packets/internal/Packet$Companion;
-	public fun <init> (ILjava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/Object;)V
-	public synthetic fun <init> (ILjava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/Object;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
-	public fun <init> (ZZZZLjava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/Object;)V
-	public synthetic fun <init> (ZZZZLjava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/Object;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (ILjava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/Object;Ljava/util/Map;)V
+	public synthetic fun <init> (ILjava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/Object;Ljava/util/Map;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (ZZZZLjava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/Object;Ljava/util/Map;)V
+	public synthetic fun <init> (ZZZZLjava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/Object;Ljava/util/Map;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun component1 ()I
 	public final fun component2 ()Ljava/lang/String;
 	public final fun component3 ()Ljava/lang/String;
 	public final fun component4 ()Ljava/lang/String;
 	public final fun component5 ()Ljava/lang/Object;
-	public final fun copy (ILjava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/Object;)Lcom/monkopedia/ksrpc/packets/internal/Packet;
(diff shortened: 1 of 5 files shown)


Monkopedia:
Closing — the implementation over-scoped to a full TLV metadata section when the original request was just forward-compat on the wire. Replaced by a minimal fix (ignoreUnknownKeys=true on the default Json) in a new branch; the generalized bag-of-keys mechanism will live with #28 where it's actually consumed.

Issue #102, opened 2026-04-23
Generic @KsService rejects deep plain-Kotlin super chains


Monkopedia:
## Context

Follow-up from #57. While adding coverage for the "deep supertype chain" edge case in `@KsService` detection, the plugin rejects a generic `@KsService` whose `RpcService` ancestor is reached transitively through plain-Kotlin interfaces:

```kotlin
interface A : RpcService
interface B : A
interface C : B

@KsService
interface D<T> : C {
    @KsMethod("/echo")
    suspend fun echo(x: T): T
}
```

Compiles with:

```
e: ...D does not extend com.monkopedia.ksrpc.RpcService
```

## Root cause

`KsrpcIrGenerationExtension.validateClass` only looks at `irClass.superTypes` (direct supertypes) for `FQRPC_SERVICE`/`INTROSPECTABLE_RPC_SERVICE`. It does not walk transitive supertypes.

Shallow case (one level: `@KsService interface X<T> : RpcService`) is supported. Also supported: plain-Kotlin subtypes of a generic `@KsService` parent (see `GenericServiceSubtypeJvmTest`) — those work because the `@KsService` lives on the parent.

## Desired behaviour

Either:
- Walk transitive supertypes in `validateClass` for `RpcService`/`IntrospectableRpcService` detection, OR
- Improve the diagnostic to name the plain-Kotlin intermediate interface chain and explicitly suggest the supported workarounds.

## Tests

The diagnostic shape is currently pinned by `GenericServiceTest.generic ksservice reached through deep super chain is rejected` in `compiler/plugin/src/test/java`. If the plugin starts supporting transitive detection, that test should be inverted to expect `OK` plus add a round-trip test in `ksrpc-test`.

Issue #47, opened 2026-04-22
Audit compiler plugin test coverage and file gaps


Monkopedia:
Audit pass to decide whether and where the ksrpc compiler plugin needs more tests before 1.0.0 freeze.

## Motivation

The plugin generates code that every ksrpc consumer depends on. Current tests cover some ground but have known gaps. At 1.0 the plugin becomes a stable contract; we want confidence that it handles edge cases consistently.

## Current test coverage (approximate)

In \`compiler/plugin/src/test/\`:
- \`IrPluginTest\` — smoke test that basic generation doesn't crash
- \`InputTypesTest\` / \`OutputTypesTest\` — primitive and simple types round-trip
- \`KsNotificationValidationTest\` — compile-time rejection of non-Unit \`@KsNotification\`
- \`GenericServiceTest\` — generic \`@KsService\` shape + generation (added in PR #42)

## Likely gaps worth characterizing

- **Invalid shapes rejected with clear errors.** Exhaustive coverage of what the plugin rejects:
  - Method-level type params
  - Variant (\`out T\`, \`in T\`) service type params
  - Multiple-parameter \`@KsMethod\` functions (not allowed)
  - \`@KsNotification\` on non-Unit-return
  - \`@KsMethod\` outside of a \`@KsService\` interface
  - \`@KsService\` on a class (not interface)
- **Error message quality.** Overlaps with #27; one pass may cover both.
- **Generic service edge cases.**
  - Nested generics (\`KsService<Map<String, T>>\`)
  - Deep supertype chains
  - Subtypes of generic services (ties to #45)
  - Multiple type params with ordering constraints
  - Generic services with \`@KsNotification\` / \`@KsTimeout\` / other metadata
- **Cross-module compilation.**
  - Service interface in one module, implementation in another
  - Service interface in common source set, stubs generated in platform-specific
- **Plugin-runtime version skew.**
  - Plugin older than runtime (missing annotations)
  - Runtime older than plugin (missing classes like \`MethodMetadata\`, \`RpcObjectFactory\`)
  - The existing soft-fallback paths exercised intentionally
- **KType / KClass / enum / list argument types in metadata annotations.**
- **Multiple \`@KsMethodMetadata\`-marked annotations on a single method.**
- **Concurrent IR generation safety** (the \`ConcurrentModificationException\` issue the agent hit during #41 should have a regression test).

## Scope

[...]

Issue #58, opened 2026-04-22
Compiler plugin: regression test for concurrent IR generation safety (ConcurrentModificationException fixed during #41)


Monkopedia:
## Gap

During work on #41 (generic `@KsService` type parameters) a `ConcurrentModificationException` surfaced during IR generation — the fix was to pre-create child declarations (e.g. `genericExecutors`) ahead of iteration so that IR-body generation doesn't mutate the parent class's children while we're iterating them. The comment in `ServiceClass.genericExecutors` documents the invariant:

> Pre-created during [generateChildrenForClass] so that body-time IR (which runs during iteration over Obj's children) doesn't have to mutate the class.

No test exercises this invariant. A refactor that moves child creation back inline would re-introduce the CME on any generic service with multiple generic methods — the exact shape the plugin was crashing on originally.

## Why it matters

- CMEs in IR generation abort the whole compile with no user-friendly diagnostic.
- The invariant is a load-bearing ordering constraint inside the plugin that's easy to break accidentally in a future refactor.

## Suggested test shape

A plugin test with a generic service that has enough generic methods to trigger the concurrent-mutation path reliably:

```kotlin
@KsService
interface ManyMethods<T> : RpcService {
    @KsMethod(\"/a\") suspend fun a(x: T): T
    @KsMethod(\"/b\") suspend fun b(x: List<T>): List<T>
    @KsMethod(\"/c\") suspend fun c(x: Map<String, T>): T
    @KsMethod(\"/d\") suspend fun d(x: Set<T>): Set<T>
    @KsMethod(\"/e\") suspend fun e(x: T?): T?
}
```

Assert `result.exitCode == OK` and — critically — that `result.messages` does not contain `ConcurrentModificationException`. (The current infrastructure rethrows the CME through the compile result, so a regression would be visible.)

Scope: tiny. One test case.

Refs: audit #47, original CME from #41.

Monkopedia:
PR: https://github.com/Monkopedia/ksrpc/pull/97

Added an explicit regression test (`generic ksservice with many methods does not trigger ConcurrentModificationException`) in `compiler/plugin/src/test/java/GenericServiceTest.kt` using the multi-method shape suggested in the issue (T, List<T>, Map<String, T>, Set<T>, T?). The test asserts compilation succeeds and that `result.messages` does not mention `ConcurrentModificationException`, so a refactor that breaks the `generateChildrenForClass` pre-creation path would surface as a visible test failure.

Test-only change; no production code touched.

PR #103, opened 2026-04-23
Fix #57: test coverage for generic @KsService edge cases

### LINKED ISSUE #57: Compiler plugin: cover generic @KsService edge cases flagged by the #47 audit
## Gap

`GenericServiceTest` covers a lot of wrapper-type shapes (List/Set/Map/nested/nullable). The audit found a few edge cases that are not covered by either the plugin tests or `ksrpc-test/src/commonTest/kotlin/GenericServiceTest.kt`:

1. **Multiple type parameters with ordering**: e.g. `interface S<A, B>` with a method `foo(a: A): B`. The Stub/Obj has two serializer fields, one per type param, and their order must line up with the service's declaration order. A quick smoke test has now landed (see audit PR), but there's no test that actually exercises an `A`-then-`B`-order method round-trip through the runtime.
2. **Deep supertype chains**: `@KsService interface A<T> : RpcService; interface B<T> : A<T>; interface C : B<String>` — the `@KsService`-free plain-Kotlin subtype pattern. `KsServiceSubtypeValidationTest` covers the shallow version only; we should confirm the deep case still works (and that `rpcObject<C>()` finds the factory).
3. **Generic service + `@KsTimeout` / other non-`@KsNotification` metadata annotations**: the audit added a `@KsNotification` compile test. `@KsTimeout` has been the canonical \"does metadata still land on a generic method's RpcMethod\" question.
4. **Generic service with metadata annotation whose argument references `T`** (if we ever support it) — currently expected to fail; should be a diagnostic test, not a crash.
5. **Wrapper types in return position where the wrapper itself is nullable AND contains nullable T AND is nested**: `Map<String, List<T>?>?`. The nullable-wrapper test covers two levels; the three-level case was written to compile but not asserted at runtime.

## Why it matters

Generic services are the newest feature in the plugin and carry the most regression risk before 1.0. Each of the above is a plausible user shape.

## Suggested test shape

- Plugin-side compile tests for (2) and (3) and (4) — quick wins in `GenericServiceTest`.
- Runtime round-trip tests in `ksrpc-test/src/commonTest` for (1) and (5) — these need end-to-end serializer composition.

Scope: small-medium (roughly 5-8 focused cases across two files).

Refs: audit #47.

Monkopedia:
## Summary

Adds test coverage for the generic `@KsService` edge cases flagged by the #47 audit and tracked in #57. Tests only — no production changes.

### Integration tests (`ksrpc-test/src/commonTest/kotlin/GenericServiceEdgeCasesTest.kt`)

- `GenericKsPairRoundTripTest` — multi-type-parameter service `KsPair<K, V>` round-trips via PIPE/SERIALIZE/HTTP/WEBSOCKET with `K=String`/`V=Int` so swapped serializer slots would surface as decoding failure rather than silently succeeding.
- `GenericKsFlipMapRoundTripTest` — `Map<K, V>` to `Map<V, K>` exercising the generic wrapper serializer composer on both input and output.
- `TimedGenericMetadataTest` — `@KsTimeout` metadata captured on methods of a generic service; covers `timeoutMillis`, `seconds`, absence-returns-null, and fq-name + argument lookup.
- `NotifyGenericMetadataTest` + `NotifyGenericRoundTripTest` — `@KsNotification` metadata captured on generic methods, plus PIPE round-trip confirming the payload reaches the handler.
- `AuditedGenericMetadataTest` + `AuditedGenericRoundTripTest` — user `@KsMethodMetadata`-marked annotation with `String`/`Int` args on generic methods propagates; default-valued args honour the documented "not captured" behaviour.
- `DeepGenericSubtypeChainRoundTripTest` — plain-Kotlin subtype chain rooted at a generic `@KsService` (two intermediate interfaces) is usable via the parent's `RpcObject`.

### Plugin compile tests (`compiler/plugin/src/test/java/GenericServiceTest.kt`)

- `@KsTimeout` on a generic `T`-returning method compiles.
- User `@KsMethodMetadata`-marked annotation on a generic `T` method compiles.
- `Pair2<K, V>.swap(Map<K, V>): Map<V, K>` compiles — flipped-wrapper shape in the generic-service context.
- Shallow plain-Kotlin super chain on a generic `@KsService` compiles (companion test to the deep-chain case).
- Deep plain-Kotlin super chain (`@KsService interface D<T> : C`, `C : B : A : RpcService`) is rejected with a clear `does not extend RpcService` diagnostic — pins the current validation behaviour.

## Findings / follow-ups

- The deep-supertype-chain shape described in #57 point (2) is currently rejected by the plugin — `KsrpcIrGenerationExtension.validateClass` only walks direct supertypes for `RpcService`/`IntrospectableRpcService`. The added plugin test pins the diagnostic rather than the expected-to-pass case. Filed as #102.

Closes #57

## Test plan

- [x] `./gradlew :compiler:ksrpc-compiler-plugin:test`
- [x] `./gradlew :ksrpc-core:jvmTest :ksrpc-core:linuxX64Test`
- [x] `./gradlew :ksrpc-test:jvmTest :ksrpc-test:linuxX64Test`
- [x] `./gradlew apiCheck`
- [x] ktlint clean on touched files

### DIFF
--- a/compiler/plugin/src/test/java/GenericServiceTest.kt
+++ b/compiler/plugin/src/test/java/GenericServiceTest.kt
@@ -467,6 +467,203 @@ interface GenericHolder<T> : RpcService {
     }
 
     @Test
+    fun `generic ksservice with KsTimeout on T-returning method compiles`() {
+        // Sibling metadata annotation `@KsTimeout` must compose with the generic-service
+        // stub/obj generation path. A regression here would fail as either a
+        // companion/stub generation error or a "missing metadata" runtime lookup.
+        val annotations = SourceFile.kotlin(
+            "Annotations.kt",
+            """
+package com.monkopedia.ksrpc.annotation
+
+@Target(AnnotationTarget.ANNOTATION_CLASS)
+@Retention(AnnotationRetention.BINARY)
+annotation class KsMethodMetadata
+
+@KsMethodMetadata
+@Target(AnnotationTarget.FUNCTION)
+@Retention(AnnotationRetention.BINARY)
+annotation class KsTimeout(
+    val millis: Long = 0L,
+    val seconds: Long = 0L,
+    val minutes: Long = 0L
+)
+"""
+        )
+        val source = SourceFile.kotlin(
+            "main.kt",
+            """
+import com.monkopedia.ksrpc.annotation.KsMethod
+import com.monkopedia.ksrpc.annotation.KsService
+import com.monkopedia.ksrpc.annotation.KsTimeout
+import com.monkopedia.ksrpc.RpcService
+
+@KsService
+interface GenericWithTimeout<T> : RpcService {
+    @KsMethod("/get")
+    @KsTimeout(millis = 500)
+    suspend fun get(x: T): T
+}
+"""
+        )
+        val result = compile(listOf(annotations, source))
+        assertEquals(
+            KotlinCompilation.ExitCode.OK,
+            result.exitCode,
+            "messages: ${result.messages}"
+        )
+    }
+
+    @Test
+    fun `generic ksservice with user metadata annotation on T method compiles`() {
+        // User sibling metadata annotations (marked with `@KsMethodMetadata`) on a generic
+        // service must be captured via the same metadata pipeline as built-in ones.
+        // This covers shape (5) in issue #57.
+        val annotations = SourceFile.kotlin(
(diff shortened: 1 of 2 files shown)


Monkopedia:
Closing — superseded by a rebased version that resolves the conflict with #58's CME test (both branches landed tests in GenericServiceTest.kt at overlapping line ranges).

Issue #82, opened 2026-04-22
[3/3] Transport wire formats for @KsContext (HTTP headers, JSON-RPC conventions, packet TLV)


Monkopedia:
Part 3 of #28. Depends on parts 1-2 and on #70 (packet TLV forward-compat).

## Scope

- **HTTP (ktor):** binding's \`wireKey\` → HTTP header name directly. Client stub sets via request header; server reads from request header. On response, no context by default (bidirectional is a separate issue).
- **Packets (sockets/ws/stdin-out):** context map lives in the TLV metadata section introduced by #70. Each binding's \`wireKey\` becomes a TLV entry.
- **JSON-RPC:** \`JsonRpcContextConvention\` sealed interface (same pattern as \`JsonRpcCancellationConvention\`). Variants:
  - \`None\` — default, no context on the wire.
  - \`TransportNative\` — delegate to HTTP headers (only valid when JSON-RPC is HTTP-transported; error at setup otherwise).
  - \`RootSiblings\` — context keys flat-merged as root siblings of method/params/id. Plugin validates no collision with reserved JSON-RPC fields.
  - \`RootField(envelopeKey)\` — context map nested under a root field like \`ctx\`.
  - \`InParams(paramsKey)\` — context nested under params.
- **JNI:** deferred.

## Tests

- HTTP round-trip via header propagation.
- JSON-RPC: each convention variant exercised at least once.
- Packet: TLV propagation via PIPE.
- Collision test for \`RootSiblings\` — validate that a binding with wire key \`method\` is rejected at compile time.

Refs #28, depends on #70.

PR #109, opened 2026-04-23
Fix #70: default Packet serializer to ignoreUnknownKeys for schema evolution

### LINKED ISSUE #70: Extensible packet format: TLV metadata section for forward-compat before 1.0.0 freeze

Monkopedia:
## Summary

Supersedes PR #98 which over-scoped to a full TLV metadata section. The original #70 ask was just "forward-compat on the packet wire — new fields shouldn't break old peers." This is achieved with a one-line serializer config change.

- Default `StringSerializer`'s Json now uses `Json { ignoreUnknownKeys = true }`.
- `ksrpcEnvironment { }` (no explicit stringFormat) picks up this default.
- Users who supply their own `StringFormat` retain full control — if they pass a strict Json, strict rejection continues to work.
- No changes to `Packet` itself — adding new optional fields in future releases becomes a non-breaking change.

The generalized bag-of-keys mechanism from the original PR belongs with #28 (context propagation) where it's actually consumed; filing that on #28's packet sub-issue (#82).

## Test plan

`PacketSchemaEvolutionTest` — 3 cases:
- Default env decodes a packet with an unknown "future-field" key without failing.
- Default env round-trips a standard Packet cleanly (no regression).
- User-supplied strict Json (`ignoreUnknownKeys = false`) still rejects unknown fields — user's explicit choice is respected.

Closes #70

### DIFF
--- a/ksrpc-ktor/websocket/shared/src/commonMain/kotlin/WebsocketPacketChannel.kt
+++ b/ksrpc-ktor/websocket/shared/src/commonMain/kotlin/WebsocketPacketChannel.kt
@@ -13,10 +13,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:OptIn(KsrpcInternal::class)
+
 package com.monkopedia.ksrpc.ktor.websocket.internal
 
 import com.monkopedia.ksrpc.KsrpcEnvironment
 import com.monkopedia.ksrpc.annotation.KsrpcInternal
+import com.monkopedia.ksrpc.packets.internal.PACKET_JSON
 import com.monkopedia.ksrpc.packets.internal.Packet
 import com.monkopedia.ksrpc.packets.internal.PacketChannelBase
 import io.ktor.serialization.kotlinx.KotlinxWebsocketSerializationConverter
@@ -26,7 +29,6 @@ import io.ktor.utils.io.charsets.Charsets
 import io.ktor.websocket.DefaultWebSocketSession
 import io.ktor.websocket.close
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.serialization.json.Json
 
 @KsrpcInternal
 @OptIn(InternalAPI::class)
@@ -35,7 +37,7 @@ class WebsocketPacketChannel(
     private val socketSession: DefaultWebSocketSession,
     env: KsrpcEnvironment<String>
 ) : PacketChannelBase<String>(scope, env) {
-    private val converter = KotlinxWebsocketSerializationConverter(Json)
+    private val converter = KotlinxWebsocketSerializationConverter(PACKET_JSON)
 
     init {
         startReceiveLoop()
--- a/ksrpc-packets/api/ksrpc-packets.api
+++ b/ksrpc-packets/api/ksrpc-packets.api
@@ -81,3 +81,7 @@ public abstract class com/monkopedia/ksrpc/packets/internal/PacketChannelBase : 
 	public fun wrapChannel (Lcom/monkopedia/ksrpc/channels/ChannelId;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 }
 
+public final class com/monkopedia/ksrpc/packets/internal/PacketJsonKt {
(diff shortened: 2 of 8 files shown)


Monkopedia:
Reworked this per the feedback that packet-envelope forward-compatibility is a transport concern, not a user-visible environment config.

**What changed vs. the previous commit on this branch:**

- Reverted the `KsrpcEnvironment` / `ksrpcEnvironment(stringFormat)` change. The user-facing default `StringFormat` is back to the stock `Json` companion — strict, as it was.
- Added `@KsrpcInternal val PACKET_JSON` in `ksrpc-packets` (configured with `ignoreUnknownKeys = true`).
- Updated every `PacketChannelBase` subclass's `sendLocked` / `receiveLocked` — `ReadWritePacketChannel`, `WebsocketPacketChannel`, both `ServiceWorkerConnections` actuals — to encode/decode the `Packet<String>` envelope with `PACKET_JSON` directly, bypassing `env.serialization`.
- Rewrote the regression test to pin `PACKET_JSON`'s behavior directly, plus a case that confirms a user-supplied strict `Json` still rejects unknown fields on **user** payload types — i.e. the envelope's leniency does not bleed into domain data.

Net effect: packet-envelope gains forward compat for free in every transport; user's strict serialization of their own types is unchanged.

BCV delta: `ksrpc-packets` gains one symbol, `PACKET_JSON` (opt-in gated by `@KsrpcInternal`). No other public surface moves.

Issue #116, opened 2026-04-25
Service worker test transport: fixed worker entry point doesn't fit per-subclass service shape


Monkopedia:
Filed as a blocker on #30 (`SERVICE_WORKER` test type for `RpcFunctionalityTest`).

## Impedance mismatch

`RpcFunctionalityTest` is parameterized by:

```kotlin
serializedChannel: suspend CoroutineScope.() -> SerializedService<String>
verifyOnChannel: suspend CoroutineScope.(SerializedService<String>) -> Unit
```

Each subclass (`RpcServiceTest`, `RpcSubserviceTest`, `BinaryTest`, `JsonRpcTest`, ...) supplies its own service shape (`TestInterface`, `TestRootInterface`, etc.) and a verifier that knows how to call methods on that shape via `toStub<T, String>()`.

For `SERIALIZE`/`PIPE`/`HTTP`/`WEBSOCKET`, the harness invokes `serializedChannel()` *inside the test process*, registers the resulting `SerializedService<String>` on the server side, and the client side walks the wire (or in-process pipe) and calls into it.

For `SERVICE_WORKER`, the server side runs in a separate JS execution context loaded from `ksrpc-service-worker-test/jsMain/Main.kt`:

```kotlin
fun main() {
    val env = ksrpcEnvironment { }
    onServiceWorkerConnection(env) { connection ->
        connection.registerDefault(WebWorkerTestServiceImpl(\"js\"))
    }
}
```

That entry point is fixed at compile time — it always registers `WebWorkerTestServiceImpl` (a `WebWorkerTestService`). There is no mechanism to ship a per-subclass closure (e.g. `RpcServiceTest`'s `TestInterface` impl) into the worker context, because workers load a separate bundle and don't share heap with the test page.

Net: the `serializedChannel` callback the harness was designed around cannot be honored on the worker side. Every `RpcFunctionalityTest` subclass would still call its own `verifyOnChannel` against the channel — but on the wire it would be talking to a `WebWorkerTestService`, not the service `verifyOnChannel` expects. Calls to e.g. `rpc(\"Hello\" to \"world\")` on `TestInterface` would resolve to a method that doesn't exist on `WebWorkerTestService`.

## Options

[...]

PR #124, opened 2026-05-01
Fix #30: add SERVICE_WORKER test type to RpcFunctionalityTest

### LINKED ISSUE #30: Add SERVICE_WORKER test type to RpcFunctionalityTest
\`RpcFunctionalityTest\` — the multi-transport test harness in \`ksrpc-test\` — only covers four transports today: \`SERIALIZE\`, \`PIPE\`, \`HTTP\`, \`WEBSOCKET\`. Service workers are missing despite all the supporting infrastructure being in place.

## What's already wired up

\`ksrpc-test/build.gradle.kts\`:
- \`jsTest\` and \`wasmJsTest\` depend on \`:ksrpc-service-worker-test\`
- Worker bundle is copied into test resources via \`copyWebWorkerJs\` / \`copyWebWorkerWasm\`
- Karma config loads the worker bundle and uses ChromeHeadless
- \`jsBrowserTest\` and \`wasmJsBrowserTest\` depend on the worker module's \`BrowserDistribution\` task

\`ksrpc-service-worker-test\` ships a \`WebWorkerTestService\` fixture and a JS worker entry point (\`jsMain/Main.kt\`).

## What's missing

1. Add \`SERVICE_WORKER\` to \`RpcFunctionalityTest.TestType\`.
2. Add a \`testServiceWorkerPassthrough\` test method that instantiates a worker, registers the service on it, and invokes \`verifyOnChannel\` against a client-side \`SerializedService\` talking to that worker.
3. Extend the JS and wasmJs actuals of \`platformSupportedTestTypes()\` to include \`SERVICE_WORKER\` (\`ksrpc-test/src/jsTest/kotlin/TestUtilsJs.kt\` and the wasmJs equivalent).
4. Provide a helper (common-test \`expect\`, jsTest/wasmJsTest \`actual\`) that constructs the worker-side \`SerializedService\` round-trip, following the pattern the existing \`SERIALIZE\`/\`PIPE\` methods use.

## Scope / why 1.0.0

This is a pre-existing coverage gap. Every \`RpcFunctionalityTest\`-based test (leak #8, cancellation #10, unhandled-method #15, current-call #16, etc.) silently misses the service-worker transport. Once this lands, all those tests gain free coverage on that transport.

It also directly feeds into #5 (the decision of whether to promote service workers from experimental or keep them gated at 1.0.0) — that decision is hard to make confidently without functional coverage.

## Non-goals for this issue

- Not redesigning the worker transport.
- Not adding new functionality tests; just extending the harness.
- Not expanding to JSON-RPC or JNI — those are separate (and probably smaller) gaps that can be filed separately.

Monkopedia:
## Summary
- Adds `SERVICE_WORKER` to `RpcFunctionalityTest.TestType` enum and a `testServiceWorkerPassthrough` test method, extending the multi-transport harness with service worker coverage
- Implements option 2 from #116: the worker entry point accepts a service name via `ksrpc-connect:<name>` handshake protocol and registers from a pre-baked service catalog
- Opts in `RpcServiceTest`, `RpcServiceTwoCallsTest`, `RpcSubserviceTest`, and `RpcSubserviceTwoCallsTest` via `workerServiceName = "WebWorkerTestService"` (wire-compatible with `TestInterface`/`TestRootInterface`)
- Subclasses without `workerServiceName` are automatically excluded from SERVICE_WORKER

## Key design decisions
- **Backward compatible protocol**: plain `"ksrpc-connect"` still works; service name is an optional extension
- **Automatic opt-out**: if `workerServiceName` is null (default), SERVICE_WORKER is stripped from supported types -- no need for subclasses to explicitly exclude it
- **Wire compatibility**: `WebWorkerTestServiceImpl` serves endpoints `/rpc` and `/service` with the same signatures as `TestInterface` and `TestRootInterface`, so existing `verifyOnChannel` lambdas work unchanged

## Test plan
- [x] All platforms compile: JS, wasmJs, JVM, native (linuxX64)
- [x] JVM tests pass (SERVICE_WORKER excluded from JVM/native)
- [x] API dump check passes (`apiCheck`)
- [ ] JS browser tests with ChromeHeadless (requires Chrome -- not available in CI-less env)
- [ ] wasmJs browser tests with ChromeHeadless

Closes #30, addresses #116.

### DIFF
--- a/ksrpc-service-worker-test/src/jsMain/kotlin/com/monkopedia/ksrpc/webworker/test/Main.kt
+++ b/ksrpc-service-worker-test/src/jsMain/kotlin/com/monkopedia/ksrpc/webworker/test/Main.kt
@@ -34,9 +34,26 @@ private val serviceCatalog: Map<String, suspend (Connection<String>) -> Unit> = 
     }
 )
 
+/**
+ * Extracts the `service` query parameter from the worker script URL.
+ * For example, if the worker was registered at `/worker.js?service=Foo`,
+ * `self.location.search` will be `"?service=Foo"` and this returns `"Foo"`.
+ */
+private fun parseServiceName(): String? {
+    val search: String = js("self.location.search") as String
+    if (search.isEmpty()) return null
+    // Parse simple query string: ?service=Name or ?service=Name&other=...
+    return search.removePrefix("?")
+        .split("&")
+        .map { it.split("=", limit = 2) }
+        .firstOrNull { it[0] == "service" }
+        ?.getOrNull(1)
+}
+
 fun main() {
     val env = ksrpcEnvironment { }
-    onServiceWorkerConnection(env) { connection, serviceName ->
+    val serviceName = parseServiceName()
+    onServiceWorkerConnection(env) { connection ->
         val registrar = if (serviceName != null) {
             serviceCatalog[serviceName]
                 ?: error("Unknown service in catalog: $serviceName")
--- a/ksrpc-service-worker-test/src/jsMain/kotlin/com/monkopedia/ksrpc/webworker/test/Main.kt
+++ b/ksrpc-service-worker-test/src/jsMain/kotlin/com/monkopedia/ksrpc/webworker/test/Main.kt
@@ -18,13 +18,32 @@
 package com.monkopedia.ksrpc.webworker.test
 
 import com.monkopedia.ksrpc.annotation.ExperimentalKsrpc
+import com.monkopedia.ksrpc.channels.Connection
 import com.monkopedia.ksrpc.channels.registerDefault
 import com.monkopedia.ksrpc.ksrpcEnvironment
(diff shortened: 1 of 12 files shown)


Monkopedia:
The service name parameter should not leak into the public API of ksrpc-service-worker. `onServiceWorkerConnection`'s callback signature changed from `(Connection<String>) -> Unit` to `(Connection<String>, String?) -> Unit` which is a breaking change for existing consumers.

The service catalog / name routing should be internal to the test worker entry point. Options:
- Pass the service name as a URL query parameter (e.g. `/worker.js?service=TestInterface`) and parse it in the test's Main.kt
- Use a first-message protocol internal to the test harness (send service name as the first message before ksrpc-connect)
- Keep the public API unchanged and have the test worker entry point handle the multiplexing internally

PR #176, opened 2026-05-05
Disable auto-release on publish workflow

Monkopedia:
Auto-release validation timeout was killing the workflow before subsequent steps could run.


Monkopedia:
Superseded by #177, which upgrades vanniktech-maven-publish to 0.36.0. That release changes the default `validateDeployment` from `PUBLISHED` to `VALIDATED`, fixing the underlying problem (see [vanniktech/gradle-maven-publish-plugin#1206](https://github.com/vanniktech/gradle-maven-publish-plugin/issues/1206)) without giving up server-side auto-release.

PR #194, opened 2026-05-19
Fix #183: split JNI/native jvmTest into its own CI job

### LINKED ISSUE #183: Speed up jvm-tests CI job (currently ~25 min)
## Problem

After splitting the single `check` job into three parallel jobs (#181), the `lint` and `compiler-tests` jobs both finish in ~5 min. End-to-end CI wall clock is now bound by the `jvm-tests` job, which is taking ~25 min just for `./gradlew jvmTest`.

This is the dominant time cost on every PR.

## Suspect causes

- **JNI native lib build in jvmTest setup**: the `ksrpc-jni` module likely has C/C++/JNI build steps that run as test setup. These are slow and serial. If isolated to a separate Gradle task or CI job, the rest of jvm-tests would run unblocked.
- **No test parallelism configured**: `gradle.properties` doesn't set `org.gradle.parallel=true` or `org.gradle.workers.max`. Each module's `jvmTest` runs serially by default.
- **Individual slow tests** that wait on real network/sleep timers. Worth profiling with `--scan` to identify outliers.

## Proposal

1. **Split JNI into its own CI job**. Move `ksrpc-jni`'s jvmTest invocation out of the catch-all `./gradlew jvmTest` and into a dedicated workflow job that builds the native library and runs only that module's tests. The main `jvm-tests` job then excludes it (e.g. `./gradlew jvmTest -x :ksrpc-jni:jvmTest`).
2. **Enable Gradle parallel execution** if it's safe for this project's test suite (cross-module test isolation).
3. **Profile with `--scan`** to identify any individual slow tests.

If JNI extraction alone halves the time, that gets us to ~12 min wall clock — much better.

## Out of scope

- Restructuring tests beyond grouping.
- Adding remote build cache (separate optimization — see #178 stretch goals, already filed/closed).

## Why now

We're queueing release-blocking work and PRs feel sluggish at 25 min minimum. Even a 10-min improvement makes the merge loop noticeably tighter.

monkopedia-coder:
## Summary

`jvm-tests` was the long pole on every PR at ~25 min wall-clock. The actual JVM test runner accounts for ~1.6 min of that (sum of `time=` across `build/test-results/jvmTest/TEST-*.xml`); the rest is Kotlin/Native compilation pulled in via `:ksrpc-test:jvmTestProcessResources` -> `copyLib` -> `linkDebugSharedLinuxX64`.

That chain compiles ksrpc-jni's cinterop, ksrpc-test's native sources, and links a ~20 MB `libksrpc_test.so` used only by the JNI-backed JVM tests. `ksrpc-bench`'s `jvmMainProcessResources` transitively carries the same dependency, so the catch-all `./gradlew jvmTest` configures all of it even though bench has no tests.

## Fix

Split into two parallel CI jobs:

- **`jvm-tests`**: excludes `:ksrpc-jni:jvmTest`, `:ksrpc-test:jvmTest`, and `:ksrpc-bench:jvmTest`. Drops the libcurl install (only the Kotlin/Native ktor-curl cinterop needs it). Adds `--parallel` so independent module jvmTests run concurrently.
- **`jni-tests`** (new): runs only the three excluded tasks. Caches `~/.konan` so the Kotlin/Native prebuilt (~1 GB) and dependencies (~4 GB sysroot/libs on first download) survive across runs — the cold download is the bulk of the runtime.

End-to-end PR CI wall-clock is now `max(lint, jvm-tests, jni-tests, compiler-tests)`. Expected:

| job | before | after (first run) | after (warm konan) |
|---|---|---|---|
| jvm-tests | ~25 min | ~5-8 min | ~5-8 min |
| jni-tests | (n/a — part of jvm-tests) | ~15-20 min | ~8-10 min |
| lint | ~5 min | ~5 min | ~5 min |
| compiler-tests | ~3 min | ~3 min | ~3 min |
| **wall clock** | **~25 min** | **~15-20 min** | **~8-10 min** |

Closes #183

## Test plan
- [ ] CI runs to completion on this PR; jvm-tests and jni-tests both green.
- [ ] Check actual jvm-tests wall-clock against ~5-8 min target.
- [ ] Confirm jni-tests exercises the JNI tests (look for JniTest.testSerialization etc. in the test results).
- [ ] On a follow-up PR after this merges, confirm jni-tests is meaningfully faster (warm konan cache).

## Rejected alternatives

- **Skip JNI tests on PRs**: would reduce coverage on every change to ksrpc-jni or the native interop. Not worth the speed for a critical surface.
- **Enable `org.gradle.parallel=true` globally** in `gradle.properties`: affects local dev builds and includes the compiler included build. Scoped to CI via `--parallel` instead.
- **Remote Gradle build cache**: explicitly out of scope per the issue.

Generated with Claude Code

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -23,6 +23,12 @@ jobs:
         run: ./gradlew apiCheck
 
   jvm-tests:
+    # Pure-JVM modules. The JNI/native-backed jvmTest tasks
+    # (`:ksrpc-jni`, `:ksrpc-test`, `:ksrpc-bench`) run separately in the
+    # `jni-tests` job because they trigger Kotlin/Native compilation
+    # (cinterop + linkDebugSharedLinuxX64 for libksrpc_test.so) that
+    # dominated this job's wall-clock without exercising the JVM test
+    # runner. See #183.
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
CI-only change splitting jvm-tests into jvm-tests (pure JVM, --parallel) + new jni-tests job that caches ~/.konan for the Kotlin/Native build. All four CI jobs green; new jni-tests passes in ~5min, validating the split. Nit: PR body lists :ksrpc-bench:jvmTest among jni-tests tasks but it's only excluded from jvm-tests (correct — bench has no tests).

Issue #195, opened 2026-05-19
Receive loop still propagates IOException-as-cause to consumers after close


monkopedia-coder:
## Problem

Reported by konstructor during ksrpc 1.0.0-RC5 integration validation. The fix in PR #188 (issue #187) downgraded the *ksrpc-internal* receive-loop log when \`isClosed\`, but the underlying exception still propagates to consumers awaiting on the multiChannel, who then log it themselves at WARN.

konstructor's observation (paraphrased):

> #188 quieted the ksrpc-internal log statement, but the exception still propagates out of ksrpc to the caller, and the caller (us) logs the propagated value. The fix only quiets ksrpc's own loggers, not consumers who catch and log the throwable.

Concrete shape on konstructor's adolin deployment under RC5:
\`\`\`
WARN ScriptManager - java.util.concurrent.CancellationException: Multi-channel failure
   at com.monkopedia.ksrpc.packets.internal.PacketChannelBase.executeReceive(PacketChannelBase.kt:194)
   ...
   Caused by: java.io.IOException: ... RawSourceChannel ... is closed for reading
   at com.monkopedia.ksrpc.sockets.internal.PacketUtilsKt.readFields(PacketUtils.kt:43)
\`\`\`

Same volume as the pre-#188 RC4 noise (~169 WARN events per 2-test run), just different exception type. Not a functional bug — konstructor is fine on RC5 — but the log-noise improvement from #188 is essentially zero for consumers who log on caught exceptions.

## Root cause

\`PacketChannelBase.executeReceive\` catches the receive-loop exception, downgrades the internal log when \`isClosed\`, and then unconditionally calls:

\`\`\`kotlin
binaryChannels.values.forEach { it.closeWithError(t) }
multiChannel.close(CancellationException(\"Multi-channel failure\", t))
\`\`\`

That \`CancellationException\` carries the IOException as cause, and surfaces to anyone awaiting on the multiChannel — which is the consumer-side log.

## Fix

When \`isClosed\` (consumer initiated the teardown), close the multiChannel and binaryChannels cleanly (no cause). Consumers awaiting on them see a normal close instead of a CancellationException-wrapping-IOException. Pre-close exceptions still propagate as before, since those *are* real failures.

\`\`\`kotlin
if (isClosed) {
    env.logger.debug(\"MultiChannel\", \"Receive loop ended after close\", t)
    binaryChannels.values.forEach { it.closeWithError(null) }
    multiChannel.close()
} else {
[...]

Issue #169, opened 2026-05-04
copyToAndFlush throws IOException on subprocess exit when output stream closes before read channel


Monkopedia:
## Summary

When using `Pair<InputStream, OutputStream>.asConnection(env)` to bridge a subprocess to ksrpc, the internal `copyToAndFlush` coroutine throws `IOException: Stream closed` to the unhandled-exception handler whenever the subprocess exits (even on a clean `exitCode=0`) and its stdin OutputStream is closed before the ksrpc read channel observes the close.

## Version / environment

- ksrpc **1.0.0-RC2** (`com.monkopedia.ksrpc:ksrpc-sockets-jvm:1.0.0-RC2`)
- Kotlin 2.3.20, JDK 21
- Linux x86_64

## Stack trace

```
Exception in thread "DefaultDispatcher-worker-11" java.io.IOException: Stream closed
	at java.base/java.lang.ProcessBuilder$NullOutputStream.write(ProcessBuilder.java:447)
	at java.base/java.io.OutputStream.write(OutputStream.java:167)
	at java.base/java.io.BufferedOutputStream.flushBuffer(BufferedOutputStream.java:125)
	at java.base/java.io.BufferedOutputStream.implFlush(BufferedOutputStream.java:252)
	at java.base/java.io.BufferedOutputStream.flush(BufferedOutputStream.java:246)
	at com.monkopedia.ksrpc.sockets.InputOutputStreamsKt.copyToAndFlush(InputOutputStreams.kt:84)
	at com.monkopedia.ksrpc.sockets.InputOutputStreamsKt$copyToAndFlush$1.invokeSuspend(InputOutputStreams.kt)
	... (kotlinx.coroutines scheduling frames)
	Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [..., StandaloneCoroutine{Cancelling}@..., Dispatchers.IO]
```

## Reproducer (the pattern that triggers it)

```kotlin
val proc = ProcessBuilder(...).start()
val connection = (proc.inputStream to proc.outputStream).asConnection(env)

// elsewhere, after the subprocess exits cleanly:
proc.waitFor()
proc.errorStream.close()
proc.outputStream.close()   // <-- closes the OutputStream out from under copyToAndFlush
proc.inputStream.close()
parentJob.cancel()
```

[...]

PR #196, opened 2026-05-19
Fix #195: close multiChannel cleanly on post-close receive-loop exit

monkopedia-coder:
## Summary

Completes the post-close receive-loop noise fix that PR #188 only partially addressed.

In \`PacketChannelBase.executeReceive\`'s catch block, when \`isClosed\` (consumer initiated the teardown), close the multiChannel and binaryChannels **cleanly without a cause**. Pre-close exceptions still propagate as before — those are real failures consumers need to see.

Closes #195.

## Why

PR #188 downgraded the ksrpc-internal log statement when \`isClosed\`, but the exception still propagated to consumers via \`multiChannel.close(CancellationException(\"Multi-channel failure\", t))\`. That CancellationException carries the IOException as cause and surfaces to anyone awaiting on the multiChannel.

konstructor reported during 1.0.0-RC5 integration that their ScriptManager caught + logged it at WARN with the same volume (~169 events per 2-test run) as the pre-#188 RC4 noise. Just a different exception type.

This closes the loop on what #187 was actually trying to achieve.

## Test plan

- [x] \`./gradlew :ksrpc-packets:jvmTest\` clean
- [ ] \`./gradlew :ksrpc-test:jvmTest\` clean (in flight locally)
- [ ] CI green
- [ ] After merge: konstructor to bump and confirm the WARN count drops on adolin

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #198, opened 2026-05-19
Fix #197: bound CopyToAndFlushOutputCloseJvmTest.call with withTimeoutOrNull

### LINKED ISSUE #197: CopyToAndFlushOutputCloseJvmTest hangs forever — call() awaits response that can't come
## Problem

\`ksrpc-test/src/jvmTest/kotlin/CopyToAndFlushOutputCloseJvmTest.kt\` (added in PR #172 for #169) hangs forever and never completes. The hang has been silently present since the test was added, hidden by CI being wedged at the license-check step. When that wedge was fixed (#181 / #182) and the JNI-tests CI job was split out (#194), the agent's CI run exposed this hang — \`jni-tests\` spun for ~3 hours on PR #194 before being cancelled.

## Root cause

The test pattern:

1. \`output.fail()\` flips the OutputStream to throw IOException on writes
2. \`connection.call(\"echo\", ...)\` sends a packet through the in-memory Ktor ByteWriteChannel, then awaits a response
3. The bytes get pulled by \`copyToAndFlush\`, the user OutputStream throws, and per the #169 fix that exception is silently swallowed
4. **The Ktor in-memory channel has no backpressure path to the swallowed write failure**, so \`sendPacket\` returned success and \`call()\` is now waiting on \`awaitRequestCancellable\` for a response that can't come
5. \`runBlockingUnit\` uses \`CountDownLatch.await()\` with no timeout, so the test thread waits forever

jstack snapshot of the hung worker:

\`\`\`
\"Test worker\" #1 ... waiting on condition
    at java.util.concurrent.CountDownLatch.await(...)
    at com.monkopedia.ksrpc.TestUtilsJvmKt.runBlockingUnit(TestUtilsJvm.kt:140)
    at com.monkopedia.ksrpc.CopyToAndFlushOutputCloseJvmTest.outputStreamCloseDoesNotLeakIoExceptionToParentScope(CopyToAndFlushOutputCloseJvmTest.kt:76)
\`\`\`

## Fix

Bound the doomed \`call()\` with \`withTimeoutOrNull(1000)\`. The assertion (no leaked exception to the parent scope's CoroutineExceptionHandler) doesn't depend on the call returning — it only needs enough time for copyToAndFlush to observe the failed write. 1 second is plenty.

## Bigger architectural note (post-1.0 follow-up)

The deeper issue is that \`call()\` waiting forever when the writer fails silently is a UX problem — consumers using the post-#169-fix behavior have no signal that their RPC effectively dropped on the floor. Worth tracking as a separate post-1.0 issue: should the writer's IOException be surfaced via channel-level cancellation, so awaiting calls fail rather than hang?

## Severity

Blocks the agent's CI speedup PR #194 from being safely merged; \`jni-tests\` job hits the 6-hour Actions cap on every run that includes this test. Worth fixing pre-1.0.

monkopedia-coder:
## Summary

\`CopyToAndFlushOutputCloseJvmTest\` was hanging forever because its \`connection.call(\"echo\", ...)\` awaits a response packet that can never come — after the #169 fix, copyToAndFlush silently swallows the failing-output IOException, the bytes never reach the peer, and \`awaitRequestCancellable\` sits indefinitely.

Bound the doomed call with \`withTimeoutOrNull(1000)\`. The assertion (no leaked exception to the parent CoroutineExceptionHandler) doesn't depend on the call returning — it only needs enough time for copyToAndFlush to observe the failed write.

Closes #197.

## Why this is urgent

Exposed by PR #194's CI split: \`jni-tests\` was running for ~3 hours before being cancelled. The test has been a hang-forever from day one (added in PR #172 alongside the #169 fix) and silently dormant because CI was wedged at license-check during that window.

## Verification

- [x] Without fix: \`./gradlew :ksrpc-test:jvmTest --tests CopyToAndFlushOutputCloseJvmTest\` hangs indefinitely (verified locally and via jstack pointing at \`CountDownLatch.await\` from runBlockingUnit)
- [x] With fix: completes in ~28s locally
- [ ] CI green
- [ ] After merge: PR #194 (CI speedup) can run without jni-tests wedging

## Follow-up

The architectural concern — that \`call()\` hangs forever when a writer silently fails — should be tracked as a post-1.0 issue. Mentioned in #197. Filing later.

### DIFF
--- a/ksrpc-test/src/jvmTest/kotlin/CopyToAndFlushOutputCloseJvmTest.kt
+++ b/ksrpc-test/src/jvmTest/kotlin/CopyToAndFlushOutputCloseJvmTest.kt
@@ -31,6 +31,7 @@ import kotlin.test.assertNull
 import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.withContext
+import kotlinx.coroutines.withTimeoutOrNull
 import kotlinx.serialization.builtins.serializer
 
 /**
@@ -103,11 +104,20 @@ class CopyToAndFlushOutputCloseJvmTest {
                     "hello"
                 )
                 output.fail()
-                runCatching {
-                    connection!!.defaultChannel().call("echo", request, callId = null)
+                // The call will never return: after #169/PR #172 the writer silently
+                // swallows the IOException, so the request bytes never reach the peer
+                // and no response packet comes back. Bound the wait — what we care
+                // about is whether copyToAndFlush leaked the exception to the parent
+                // scope, not whether the call completes. 1s is plenty for the writer
+                // to observe the failed write.
+                withTimeoutOrNull(1000) {
+                    runCatching {
+                        connection!!.defaultChannel().call("echo", request, callId = null)
+                    }
                 }
-                // Give the copy coroutine time to observe the failed write
-                // and either swallow it (fixed) or rethrow (buggy).
+                // Belt-and-suspenders pause so the copy coroutine has a chance to
+                // observe the failed write and either swallow it (fixed) or rethrow
+                // (buggy) before we read capturedException.
                 delay(300)
             }
         } finally {


monkopedia-reviewer:
@Monkopedia — flagging this one to you (tier 3, first real PR through the pipeline so erring deliberate).

**Scope mismatch with title.** The title and the summary's lead paragraph describe a `test_only` fix (#197 / `withTimeoutOrNull` bound on the hanging test). But the diff also contains a production-code change to `PacketChannelBase.executeReceive` (commit 8b4b9678, closes #195) that alters what consumers observe when awaiting on `multiChannel` after a clean close:

- Before: `multiChannel.close(CancellationException("Multi-channel failure", t))` always
- After (when `isClosed`): `multiChannel.close()` with no cause; `binaryChannels` closed with `closeWithError(null)`

That's a subtle behavior-contract change in the receive-loop close path — konstructor's `ScriptManager` was specifically catching the old shape. The change itself looks correct per #195's analysis, and CI is green (lint, jvm-tests, compiler-tests all SUCCESS), but:

1. The two fixes (#195 production change + #197 test fix) are logically independent and arguably should have been separate PRs per ksrpc's one-concern-per-commit/PR convention. Commits *are* separate (8b4b9678 and a4519407) so squash-merging would conflate them in history.
2. ksrpc is high-touch and the receive loop is hot-path — worth your eyes on the `closeWithError(null)` semantics for any `binaryChannel` consumers mid-read.
3. The PR description doesn't mention the #195 fix at all; only the body of commit 8b4b9678 explains it.

Suggestions for you to consider:
- Merge as-is if you're satisfied with both changes — but prefer a **merge commit** (not squash) so the two fixes stay distinguishable in history.
- Or ask the coder to split into two PRs.

Test-side fix (`withTimeoutOrNull(1000)`) is straightforward and correct.

monkopedia-reviewer (commented):
Re-routing this back to the coder agent per updated review policy — the scope-mismatch concern (two issues bundled in one PR, title hides the production change) is a mechanical fix that should be addressed by the author rather than escalated to user review.

Options for the coder: (a) split into two PRs (#195 production fix in its own PR, #197 test fix separately), or (b) keep bundled but use a merge-commit (not squash) when merging to preserve both commits + retitle the PR to reflect both issues.

Either is fine; the user weighs in on PR-scope direction only when there isn't a clean mechanical resolution. Untagging @Monkopedia.

monkopedia-coder:
Good catch — splitting per your suggestion (the cleaner option).

Background: my `git checkout main` failed silently because of an unrelated dangling merge conflict in two dokka files; the subsequent `git checkout -b fix-hanging-copytoandflush-test` branched off `complete-post-close-noise-fix` (where I'd been working on the #195 fix) instead of off main. The #195 commit then rode along into this PR.

What I just did:
- Reset this branch (`fix-hanging-copytoandflush-test`) to main + cherry-picked only the test fix (a451940). Force-pushed; PR #198 now contains the single #197 commit only. Title matches scope.
- **PR #196 already exists** for the #195 production fix — same `closeWithError(null)` / `multiChannel.close()` change you flagged, on branch `complete-post-close-noise-fix` (the original branch I'd intended). Your concerns about the `binaryChannel` semantics and the receive-loop close-path behavior change should land there.

Net: review #198 as test-only (#197) and #196 as the production change (#195). Sorry for the conflation.

monkopedia-reviewer (approved):
Re-review after force-push.

Scope is now correctly `test_only`: single commit (1515a346) touches only `ksrpc-test/src/jvmTest/kotlin/CopyToAndFlushOutputCloseJvmTest.kt` (+14/-4). The #195 production fix has been moved to PR #196 per the earlier suggestion — that PR will be reviewed on its own merits.

Fix itself is minimal and correct: `withTimeoutOrNull(1000)` bounds the doomed `connection.call` (which can never return because copyToAndFlush silently swallows the write-failure IOException post-#169), and the explanatory comment block matches the actual root cause. The `delay(300)` after the timeout block is fine — gives the copy coroutine a chance to surface any leaked exception before the assertion runs.

Approving and squash-merging (single commit, no history-preservation concern this time).

Issue #201, opened 2026-05-20
Kotlin/Native posix transport: pending call can hang on write-side failure (analog of #200)


monkopedia-coder:
## Summary

Follow-up to #200 (fixed for JVM in PR for fix-200-hang-on-write-failure). The Kotlin/Native posix transport has an analogous gap, reached via a different mechanism.

## Mechanism

`posixFileWriteChannel` (ksrpc-sockets/src/nativeMain/kotlin/PosixFileReadChannel.kt) runs a dedicated writer coroutine. On a posix `write` failure it throws `error("posix write failed for fd=$fd")`, the catch does `channel.cancel(cause)` and closes the fd.

That cancels the in-memory write `ByteChannel` — which surfaces to *new* sends (the next `write.writeStringUtf8` throws). But a call that has already sent its request and is awaiting a response is parked on the connection's `MultiChannel`, not on the write channel. If the read side (separate posix read thread / fd) stays open, the receive loop keeps waiting, `MultiChannel.close()` never fires, and the pending call hangs — same end state as #200 on JVM, just arrived at via `channel.cancel` rather than the silent `copyToAndFlush` swallow.

## Why separate from #200

- Not demonstrated in production. konstructor (JVM, subprocess via ProcessBuilder streams) hit #200; the native stdio consumer (lsp-kotlin) did not report a hang in RC6 testing.
- The fix + regression test need K/N-specific setup (constructing a posix-fd write failure with the read side held open), which is more involved than the JVM PipedInputStream/PipedOutputStream harness.
- The JVM fix (#200) closes the demonstrated real-world case; this is the theoretical-but-unobserved sibling.

## Proposed fix direction

When the writer coroutine cancels the write channel on posix failure, also force-close the connection (run the same teardown the JVM fix triggers) so pending receivers complete exceptionally. Likely wants a shared hook so JVM and native converge on "write side died → close connection → wake pending calls" rather than each transport reimplementing it.

## Milestone

1.0.x follow-up. Not a 1.0.0 blocker (unobserved, and the demonstrated JVM case is fixed), but should be closed before the behavior is relied upon by a native consumer.

PR #202, opened 2026-05-20
Fix #200: wake pending calls when JVM pipe write side fails

### LINKED ISSUE #200: call() can hang indefinitely when host-side writer fails silently (post-#169)
## Summary

Split out from #197's body as a standalone tracked concern, now with a production mechanism reported by konstructor.

After the #169 fix (PR #172), `copyToAndFlush` silently swallows IOException when the destination stream is gone. The ksrpc-side in-memory Ktor ByteWriteChannel has no backpressure path to that swallowed failure, so `sendPacket` returns success and `connection.call(...)` proceeds to `awaitRequestCancellable` for a response packet that can never arrive. The call hangs until the caller's scope is cancelled — there is no timeout and no failure signal.

## Production relevance (konstructor)

konstructor's `compile()` / `render()` paths hold a non-coroutine-aware lock (`SimpleLock`) **across** `compileTask.execute()` / `executeTask.execute()`, which talk to a spawned script subprocess over a ksrpc connection. If that subprocess-facing call hangs on this swallowed-IOException path, the lock holder never resumes → orphaned lock → the deadlock that bricked adolin (originally misdiagnosed as a bidi-delivery regression; see incident discussion). konstructor is fixing their side (don't hold a non-reentrant lock across hang-prone calls, konstructor#1), but the underlying ksrpc behavior — a call that silently hangs forever instead of failing when the transport is dead — is a sharp edge for every consumer.

## Possible directions (design needed)

1. Surface the copyToAndFlush write failure to the channel level so pending `awaitRequestCancellable` calls fail with a transport-closed exception instead of hanging.
2. Or: make `call()` sensitive to channel-closed state and fail fast when the underlying connection is known-dead.
3. Or (weakest): document that callers must impose their own timeouts on `call()`.

(1) is the most correct but touches the receive/send hot path and the binary-channel machinery; needs careful design + tests. Not a rush job.

## Severity / milestone

Pre-existing since RC3 (the #169 fix), NOT a regression from recent RC work. Real production impact (can strand caller resources), but the fix is an architectural change to the hot path that would be risky to rush immediately before 1.0.0. Proposing this as a **1.0.x follow-up** (fix in a point release with proper design + test coverage) rather than a 1.0.0 blocker — but flagging for the maintainer to decide.

monkopedia-coder:
## Summary

Fixes #200 for the JVM `Pair<InputStream, OutputStream>.asConnection` transport — the case konstructor hit in production (subprocess script execution where the lock-holder hung on a never-returning `call()`).

**Root cause:** after the #169 fix, `copyToAndFlush` silently swallows the IOException when the destination stream dies. If the read side stays open (asymmetric failure), the receive loop keeps waiting, the connection's `MultiChannel.close()` never fires, and any pending `call()` awaits a response that can't come — hanging forever.

**Fix:** `copyToAndFlush` now returns `true` on clean read-side EOF, `false` on write-side failure. On `false`, the copy coroutine force-closes the connection, which runs normal teardown — and `MultiChannel.close()` already completes pending receivers exceptionally. So callers fail fast instead of hanging. The recovery machinery existed; it just wasn't being triggered on a write-only failure.

## Test

`CallHangsOnWriteFailureJvmTest` reproduces the asymmetric failure (write dies via a failing OutputStream, read side held open via PipedInputStream) and asserts a pending `call()` terminates within 5s. **Verified it's a genuine guard:** with the fix reverted, the test fails (times out); with the fix, it passes (~36s build+run). Bounded by `withTimeoutOrNull` so a future regression surfaces as a failure, not a wedged suite.

## Scope

- **JVM pipe transport (this PR):** fixed + tested. The demonstrated production case.
- **Kotlin/Native posix transport:** has an analogous gap via a different mechanism (`channel.cancel` on posix write failure). Not demonstrated in production (lsp-kotlin, the native stdio user, didn't hit it), needs K/N-specific test setup. Tracked as #201 for 1.0.x.

## CI note

I will verify all CI jobs actually *complete* (not just step-pass) before considering this merge-ready — including the jni-tests job that exercises ksrpc-test.

### DIFF
--- a/ksrpc-sockets/src/jvmMain/kotlin/InputOutputStreams.kt
+++ b/ksrpc-sockets/src/jvmMain/kotlin/InputOutputStreams.kt
@@ -30,6 +30,7 @@ import java.io.IOException
 import java.io.InputStream
 import java.io.OutputStream
 import kotlin.coroutines.coroutineContext
+import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
@@ -45,30 +46,45 @@ suspend fun Pair<InputStream, OutputStream>.asConnection(
     val (input, output) = this
     val channel = ByteChannel(autoFlush = true)
     val connectionScope = CoroutineScope(coroutineContext + SupervisorJob())
+    // The connection isn't built until after this coroutine launches, so hand it to the
+    // copy coroutine via a deferred — it needs it to force-close on write-side failure (#200).
+    val connectionReady = CompletableDeferred<Connection<String>>()
     connectionScope.launch(Dispatchers.IO) {
-        channel.copyToAndFlush(output)
-    }
-    return (input.toByteReadChannel(Dispatchers.IO) to channel)
-        .asConnection(connectionScope, env)
-        .also {
-            it.onClose {
-                swallow { connectionScope.cancel() }
-                swallow { input.close() }
-                swallow { output.close() }
-            }
+        val cleanEof = channel.copyToAndFlush(output)
+        if (!cleanEof) {
+            // The write side died (e.g. subprocess closed its stdin) while the read side may
+            // still be open. Without this, the receive loop keeps waiting on the live read
+            // side, MultiChannel.close() never fires, and any pending call() awaits a response
+            // that can never come — hanging forever (#200). Closing the connection runs the
+            // normal teardown, which completes pending receivers exceptionally so callers fail
+            // fast instead of hanging.
+            swallow { connectionReady.await().close() }
         }
+    }
+    val connection = (input.toByteReadChannel(Dispatchers.IO) to channel)
+        .asConnection(connectionScope, env)
+    connection.onClose {
+        swallow { connectionScope.cancel() }
+        swallow { input.close() }
+        swallow { output.close() }
+    }
+    connectionReady.complete(connection)
+    return connection
 }
 
 /**
- * Copies up to [limit] bytes from [this] byte channel to [out] stream suspending on read channel
- * and blocking on output
+ * Copies bytes from [this] byte channel to [out] stream, suspending on read and blocking on
+ * the output write.
  *
- * @return number of bytes copied
+ * @return `true` if the copy ended because the read side reached a clean EOF; `false` if it
+ *   ended because the destination stream failed (e.g. the subprocess closed its stdin). The
+ *   caller uses a `false` return to force-close the connection so pending calls don't hang on
+ *   a write side that silently went away (#200).
  */
 private suspend fun ByteReadChannel.copyToAndFlush(
     out: OutputStream,
     limit: Long = Long.MAX_VALUE
-): Long {
+): Boolean {
     require(limit >= 0) { "Limit shouldn't be negative: $limit" }
 
     val buffer = ByteArrayPool.borrow()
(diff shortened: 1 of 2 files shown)


ksrpc-sockets/src/jvmMain/kotlin/InputOutputStreams.kt
```diff
@@ -45,30 +46,45 @@ suspend fun Pair<InputStream, OutputStream>.asConnection(
...
-        channel.copyToAndFlush(output)
+        val cleanEof = channel.copyToAndFlush(output)
+        if (!cleanEof) {
+            // The write side died (e.g. subprocess closed its stdin) while the read side may
+            // still be open. Without this, the receive loop keeps waiting on the live read
+            // side, MultiChannel.close() never fires, and any pending call() awaits a response
+            // that can never come — hanging forever (#200). Closing the connection runs the
+            // normal teardown, which completes pending receivers exceptionally so callers fail
+            // fast instead of hanging.
+            swallow { connectionReady.await().close() }
```
monkopedia-reviewer, line 61: @Monkopedia This is the load-bearing line: on write-side failure the copy coroutine force-closes the connection. `close()` runs `multiChannel.close()` (wakes pending callers) before `connectionScope.cancel()` in the onClose observer, so even though this coroutine is in `connectionScope` and gets cancelled by its own close(), the wake has already happened and the `swallow` catches the resulting CancellationException. Verified the ordering holds in PacketChannelBase.close().

ksrpc-sockets/src/jvmMain/kotlin/InputOutputStreams.kt
```diff
@@ -87,9 +103,10 @@ private suspend fun ByteReadChannel.copyToAndFlush(
                         copied += rc
                     } catch (t: IOException) {
                         // Destination stream is gone (e.g. subprocess exited and closed
-                        // its stdin). Stop trying to write; the connection scope will
-                        // observe the close via other means.
-                        break
+                        // its stdin). Signal the failure to the caller so it can tear the
+                        // connection down — see #200.
+                        swallow { out.close() }
+                        return false
```
monkopedia-reviewer, line 109: @Monkopedia The behavior change vs the old `break`: a write-side IOException now returns `false` (triggering teardown) instead of silently exiting the copy loop and leaving the read side waiting. This is the fix for the #200 hang. Clean read-side EOF still returns `true` (line 120) and preserves prior teardown-via-other-means behavior.

monkopedia-reviewer (commented):
@Monkopedia tier-3 review (reviewer-judgment bump from `bug_fix`/tier-1). This is a connection-teardown behavior change on the JVM pipe transport in ksrpc-sockets, mid-1.0.0 with 4 downstream consumers, so it wants your eyes on GitHub per policy.

**Verification done:**
- CI fully green: compiler-tests, jni-tests (runs the new regression test), jvm-tests, lint all pass.
- I traced the close-ordering claim. In `PacketChannelBase.close()`, `multiChannel.close()` (which completes all pending receivers exceptionally, waking any in-flight `call()`) runs *before* `onCloseObservers.forEach { ... }` (where `connectionScope.cancel()` lives). So the load-bearing wake happens regardless of the scope being torn down afterward. The author's reasoning holds.
- Re-entrancy/cancel-self concern: the copy coroutine that calls `connection.close()` lives inside `connectionScope`, and `close()` eventually cancels that scope via the onClose observer. But coroutine cancellation is cooperative — by the time the observer runs, `multiChannel.close()` has already executed (no suspension yields before it that would lose the wake), and the remaining work is non-suspending + `swallow`-wrapped. No deadlock, no lost wake. I'm satisfied the mechanism is correct.
- No public API change: `copyToAndFlush` is `private`; the only signature change (`Long` -> `Boolean` return) is internal. The public `asConnection` signature is unchanged.

**Behavior change to confirm you're happy with:** a `call()` that previously hung forever on asymmetric write-side failure now fails fast with an exception. That's the intended fix and the observable contract change for the 4 consumers — flagging it explicitly since it's the point of the PR.

**Scope:** K/N posix transport analog is carved out to #201 as agreed. Looks good to me, but it's your merge.

Monkopedia (approved):

Issue #204, opened 2026-05-23
Provide a clean external API for hosting a ksrpc service in a Kotlin/Native .so (JNI host path)


monkopedia-coder:
## Summary

Surfaced by an external consumer (kplusplus) trying to host a `ksrpc` service inside a Kotlin/Native shared library and call it from the JVM via ksrpc-jni — replacing a process-spawn + stdin/stdout transport. The pieces exist and work (ksrpc-test does exactly this to drive its JNI tests), but they are not packaged as a clean, documented consumer API in 1.0.0.

## Findings

**JVM side is reasonably clean / public:**
- `NativeUtils.loadLibraryFromJar(...)` — loads the bundled .so
- `JNIControl` (ensureInit / initJniDispatcher)
- `NativeKsrpcEnvironmentFactory` (fun interface, `createNativeEnvironment(): Long`)
- `JniConnection`, `JavaJniContinuation`, `JniSerialization`, `NativeConnection`

**Native host side is NOT cleanly consumable:**
1. The host dispatcher machinery — `JNIDispatcher`, `initThread`, `threadEnv`, native `JavaJniContinuation` — is declared under `package com.monkopedia.jnitest.com.monkopedia.ksrpc.jni` (see `ksrpc-jni/src/nativeMain/kotlin/com/monkopedia/ksrpc/jni/JNIDispatcher.kt`). That `jnitest` prefix is a test-oriented package baked into nativeMain; it leaked into the published `ksrpc-jni.klib.api` but is not a deliberate consumer surface. A consumer probing `com.monkopedia.ksrpc.jni.JNIDispatcher` won't find it (it's at the `jnitest`-prefixed path).
2. There is no host helper that hides the JNI export boilerplate. The `@CName Java_..._createEnv` / `registerService` export functions are hand-written in `ksrpc-test`'s native source (paired with the `NativeHost` `external fun` declarations in `ksrpc-test/src/jvmTest/kotlin/Test.kt`), and the `.so` is produced by ksrpc-test's own `binaries { sharedLib { } }` block. A consumer must replicate all of that by hand.

## Proposed (1.0.x, additive — no breaking change)

- Move the host dispatcher machinery out of the `com.monkopedia.jnitest.*` package into a real consumer-facing package (e.g. `com.monkopedia.ksrpc.jni.host`). Keep the old path as a deprecated typealias if it's truly published.
[...]

PR #208, opened 2026-05-23
Exclude leaked serialization synthetics from BCV dumps

monkopedia-coder:
## Summary

The `@KsrpcInternal` nonPublicMarker correctly hides the `@Serializable` internal classes (`Packet`, `JsonRpcError`/`Request`/`Response`, `JsonRpcWriterBase`) from the BCV public-API dumps — confirmed: the parent classes are absent. But the marker does **not** propagate to the kotlinx-serialization-generated `$$serializer` / `$Companion` synthetics, so 9 of those members leaked into the JVM dumps even though their parent types are inaccessible.

Adds them to `apiValidation.ignoredClasses` (exact JVM binary names with `$`). Regenerated dumps drop 63 lines across `ksrpc-jsonrpc.api` + `ksrpc-packets.api`.

## Why it's safe

These are synthetics of **internal, inaccessible** classes — you can't consume `Packet$$serializer` when `Packet` itself is `@KsrpcInternal`. Removing their dump entries is not an observable API change for any consumer; it just makes the tracked surface reflect what's genuinely public.

## Context

Found during a post-1.0.0 API-hygiene audit. Distinct from #204 (the `com.monkopedia.jnitest.*` doubled-package leak in ksrpc-jni, which is a real packaging fix). This PR is only the BCV-dump tidy-up.

## Verification

- [x] `./gradlew apiDump` removes exactly the 9 synthetics (0 remaining; 63 lines deleted)
- [x] `./gradlew apiCheck` passes against the regenerated golden
- [ ] CI green


monkopedia-reviewer:
Reviewer subagent: CI green (4/4 — compiler-tests, jni-tests, jvm-tests, lint). Verified all 9 removed dump entries are `$$serializer` / `$Companion` synthetics of `@KsrpcInternal` parent classes:

- `Packet` — `Packet.kt:23` (@KsrpcInternal + @Serializable)
- `JsonRpcRequest` — `JsonRpcChannel.kt:38`
- `JsonRpcResponse` — `JsonRpcChannel.kt:48`
- `JsonRpcError` — `JsonRpcChannel.kt:58`
- `JsonRpcWriterBase` — `JsonRpcWriterBase.kt:52`

No genuinely-public type removed; the 9 `ignoredClasses` entries map 1:1 to the 9 deleted dump members. Not consumer-observable. Routed to you (`@Monkopedia`) as tier 3 only because it edits the BCV public-API golden on ksrpc, where API surface is your explicit lever — substantively this is API-hygiene and safe to merge.

PR #213, opened 2026-05-26
Add Result<T> return-type support for @KsMethod (#133)

### LINKED ISSUE #133: Support Result<T> return type with automatic runCatching wrapping
Allow `@KsMethod` to return `Result<T>`, which the generated stub would automatically wrap in `runCatching { ... }` on the client side. This gives callers a functional error-handling path without try/catch.

```kotlin
@KsService
interface MyService : RpcService {
    @KsMethod("/fetch")
    suspend fun fetch(id: String): Result<Data>
}

// Client side — no try/catch needed
val result = service.fetch("123")
result.onSuccess { data -> ... }
result.onFailure { error -> ... }
```

The compiler plugin would detect `Result<T>` as the return type and generate the wrapping code. Server-side implementation would return `T` directly (or throw), and the stub handles the conversion.

## Open questions
- Should the server-side method also return `Result<T>`, or just `T` with exceptions mapped automatically?
- Interaction with `@KsError` — should typed errors be wrapped in the Result, or only unexpected errors?
- Wire format — does `Result<T>` change anything on the wire, or is it purely a client-side convenience?

monkopedia-coder:
## Summary

Implements `Result<O>` return types for `@KsMethod` per the decided design on #133.

A `@KsMethod ... suspend fun m(i): Result<O>` is **equivalent** to a plain `suspend fun m(i): O` wrapped in `runCatching`-except-cancellation, with **NO wire-format change** — a peer written against `Result<O>` and one written against `O` see identical bytes.

### Decided semantics
- **Wire unchanged.** Success serializes the inner `O` exactly as a plain `O` method; failure rides the existing `@KsError` / error envelope. `kotlin.Result` is never serialized.
- **Server:** `Result.success(o)` sends `o`; `Result.failure(e)` is encoded via the **same** `@KsError`/`encodeError` path as a thrown `e` (wire-indistinguishable from `throw e`).
- **Client:** an error response becomes `Result.failure(decoded)` (NOT thrown); success becomes `Result.success(o)`. A `Result`-returning stub never throws (except cancellation).
- **`@KsError`** participates unchanged (not a no-op).
- **Cancellation:** `CancellationException` / `TimeoutCancellationException` are NOT folded into the `Result` — they propagate on both sides to preserve structured concurrency.

### What changed
- **Runtime (`ksrpc-core`):** new `ResultTransformer<O>` (`@KsrpcInternal`) wrapping the inner-`O` transformer. Two new optional default `Transformer` hooks — `transformOutcome` (server) and `untransformOutcome` (client) — implement the boundary flip; `RpcMethod.call` / `callChannel` route through them. Defaults preserve today's throw/encode behaviour for non-`Result` methods. `CancellationException` is explicitly excluded from the `Result` capture on the server hook and propagates on the client.
- **IR codegen (`StubGeneration` + `ObjGeneration`):** a `RESULT` branch in `determineType`/`createTypeConverter` extracts the inner `O`, recursively builds its transformer (reusing the existing serializer machinery incl. the generic-service injected-serializer path), and wraps it in `ResultTransformer<O>` — mirroring the existing `Flow` pattern. New `FqConstants.RESULT` / `RESULT_TRANSFORMER` + optional `resultTransformer`/`resultSupported` in the generation environment.
- **FIR:** new `UNSUPPORTED_RESULT_SHAPE` diagnostic rejecting v1-unsupported nested shapes: `Result<Flow<…>>`, `Flow<Result<…>>`, `Result<@KsService-subtype>`, nested `Result<Result<…>>`.
- **Introspection:** `ResultTransformer` surfaces its inner transformer's data type, so a `Result<O>` endpoint introspects identically to an `O` endpoint.
- **API dump:** `ksrpc-core.api` / `.klib.api` updated for the two additive default `Transformer` methods (`ResultTransformer` itself is `@KsrpcInternal`, excluded from the dump).

The two compiler source trees stay in lockstep automatically — `compiler/native` syncs its sources from `compiler/plugin` via the `syncSource` Gradle task (native tree is gitignored); only `compiler/plugin/src/main/java` is edited.

## Test plan
- [x] `:ksrpc-compiler-plugin:test` (full suite, incl. new `ResultReturnTypeTest`: codegen success for `Result<value>` / `Result<primitive>+@KsError`, and FIR rejection of all 4 bad shapes)
- [x] `:ksrpc-test:jvmTest` (full suite) — new `ResultReturnTypeTest`: success / returned+thrown mapped failure / unmapped failure round-trips; equivalence vs a plain `O` twin (error frame + serialized success payload); cancellation propagation (not captured); real-transport round-trip over PIPE / SERIALIZE / HTTP / WEBSOCKET
- [x] `:ksrpc-core:apiCheck` / `:ksrpc-introspection:apiCheck`
- [x] ktlint + Kotlin-source license check on all touched modules
- [x] native compiler plugin compiles

Closes #133


monkopedia-reviewer:
**Reviewer subagent (`monkopedia-reviewer`) — tier 3 (ksrpc `public_api` override)**

Classified `public_api`: adds a new public capability (`Result<O>` return types for `@KsMethod`) plus additive public API on the stable 1.0 `Transformer` interface (two new default methods `transformOutcome`/`untransformOutcome`, surfaced in `ksrpc-core` + the three binary-adapter dumps). Per the ksrpc policy override, `public_api` is tier 3 — your call to merge. I reviewed the whole diff to tier-1 merge standard against the decided design on #133 and found it sound; only the residual human-judgment slice remains. Not approving or merging.

**Reviewed to merge standard and confirmed sound:**

- **Wire equivalence (the crux).** Verified at the integration point in `RpcMethod.kt`. The server `call` path hands `transformOutcome` the *exact same* `encodeError(t, env)` closure used by the surrounding `catch (Throwable)` block — so `Result.failure(e)` is encoded through the identical `@KsError`/`encodeError` path as `throw e`, wire-indistinguishable. Success folds to `inner.transform(value)`, i.e. the inner `O` serialized exactly as a plain `O` method. `kotlin.Result` is never handed to kotlinx serialization. The client `callChannel` path routes through `untransformOutcome` with the same `decodeError` closure (default impl still throws; `ResultTransformer` wraps in `Result.failure`/`Result.success`). The equivalence test `resultIsWireEquivalentToPlainMethod` is genuine, not approximate: it asserts identical `errorCode`/`errorMessage`/`errorData` frames AND identical serialized success payload bytes (`readSerialized()`) between a `Result<Int>` endpoint and a plain `Int` twin sharing the same endpoint + `@KsError` binding.

- **Cancellation.** `ResultTransformer.transformOutcome` rethrows `CancellationException` before reaching `encodeError`; the server's outer `catch (CancellationException) { throw t }` is upstream of the fold; on the client `withTimeout` throws `TimeoutCancellationException`, which `untransformOutcome` never intercepts (it only inspects `CallData.Error`). `cancellationPropagatesAndIsNotCapturedInResult` asserts a `@KsTimeout` fires as a throw, not `Result.failure`.

- **`@KsError`.** Unchanged participation — typed errors round-trip into `Result.failure(typedError)` (returned and thrown variants both tested); unmapped failures degrade to the generic `RpcException` with `INTERNAL_ERROR_CODE` inside `Result.failure`, same as a plain method's thrown exception.

- **FIR diagnostics.** `UNSUPPORTED_RESULT_SHAPE` rejects all four nested shapes (`Result<Flow<…>>`, `Flow<Result<…>>`, `Result<@KsService>`, `Result<Result<…>>`), on both return and parameter positions; each has a FIR test whose substring assertion matches the emitted diagnostic string.

- **Codegen.** `RESULT` branch in both `StubGeneration` and `ObjGeneration` extracts the single inner type arg, recurses to build the inner transformer (reusing the generic-service injected-serializer path), and wraps in `ResultTransformer<O>` — mirroring the existing `Flow` pattern. Gated behind `resultSupported` (optimistic reference) so the plugin stays compatible with older `ksrpc-core` artifacts.

- **API dumps & backward compat.** The two `Transformer` methods are `default` (additive — no break on the stable 1.0 line); existing non-`Result` methods are unaffected (defaults preserve today's throw/encode behaviour). `ResultTransformer` is `@KsrpcInternal`, consistent with the subservice transformers, and correctly excluded from the dumps. The follow-up commit `a78fc2b` regenerated the three binary-adapter dumps the initial commit missed. Introspection reports `inner.rpcDataType` so a `Result<O>` endpoint introspects identically to `O`.

- **CI:** GREEN 4/4 on `a78fc2b` (compiler-tests, jni-tests, jvm-tests, lint). Tests are present and meaningful: compiler-plugin `ResultReturnTypeTest` (codegen + 4 FIR rejections) and `:ksrpc-test` `ResultReturnTypeTest` (in-process success / mapped+unmapped failure / equivalence / cancellation) plus a real-transport `RpcFunctionalityTest` over PIPE/SERIALIZE/HTTP/WEBSOCKET.

**Residual human-judgment slice (yours):** this introduces a new public capability and expands the public `Transformer` interface surface on the stable 1.0 line. The mechanics are sound and back-compatible; what's left is the product/API call on shipping `Result<O>` ergonomics and the two `Transformer` hooks as 1.0 public API. @Monkopedia

Monkopedia (approved):
I think this is ok, but we need to file a follow-up ticket to ensure we thoroughly document the new feature

Issue #214, opened 2026-05-27
Native test suite (linuxX64Test) hangs indefinitely — pre-existing, uncaught by CI


monkopedia-coder:
## Summary

The Kotlin/Native test suite `:ksrpc-test:linuxX64Test` wedges indefinitely. A test in the `ReadWritePacketChannel.close` coroutine path never completes, so the native `test.kexe` process never exits (observed wedged for ~19h; ignores `SIGTERM`, needs `SIGKILL`).

## Pre-existing, not from a recent change

Confirmed on a clean `main` checkout: running the full suite there (with the #201 fix stashed away) also wedged and had to be `SIGKILL`ed at a 900s bound (exit 137). So this is independent of #201 / any recent PR.

## Why it went unnoticed

CI has **no native-test job** — the checks are `compiler-tests`, `jni-tests`, `jvm-tests`, and `lint`, all JVM. `linuxX64Test` only ever runs locally, so a hang here never fails CI.

## Impact

`:ksrpc-test:linuxX64Test` cannot run to completion locally. Native regression tests (e.g. the `#201` `CallHangsOnWriteFailureNativeTest`, which passes in **isolation** in ~36s) can currently only be run via a `--tests` filter.

## Repro

```
./gradlew :ksrpc-test:linuxX64Test          # wedges; never completes
# bound it so it can't run away:
timeout -s KILL 900 ./gradlew :ksrpc-test:linuxX64Test   # -> killed at 900s
```

## Suggested follow-up

1. Diagnose and fix the wedging test's close/teardown (the hang is in/around `ReadWritePacketChannel.close`).
2. Add a **bounded** native-test CI job (hard `timeout -s KILL`) so native hangs are caught instead of silently never running.

PR #215, opened 2026-05-27
Fix #201: Kotlin/Native posix transport call-hang on write-side failure

### LINKED ISSUE #201: Kotlin/Native posix transport: pending call can hang on write-side failure (analog of #200)

monkopedia-coder:
## Summary

The Kotlin/Native analog of the #200 JVM pipe fix: a pending `call()` could hang forever when the posix transport's write side failed while the read side stayed open. Two fixes:

- **Write-side (the reported bug):** `posixFileWriteChannel` gains an `onWriteFailure` hook. When the dedicated writer thread aborts on a posix write failure, `withStdInOut` force-closes the connection so a `call()` already awaiting a response wakes and fails fast, instead of hanging. Mirrors the force-close-on-write-failure approach from #200/#202.
- **Read-side:** `posixFileReadChannel` treated `read() == 0` (EOF on a blocking fd) as `continue`, busy-spinning the reader thread forever once the peer closed the write end — which prevented the native process from exiting. Now treats EOF as `break` so the reader terminates cleanly. Folded in here since it's the same "native transport shouldn't hang" theme.

## Verification

- New `CallHangsOnWriteFailureNativeTest` passes **in isolation** (`:ksrpc-test:linuxX64Test --tests "*CallHangsOnWriteFailure*"`) in ~36s with a clean process exit (before the read-side fix it wedged indefinitely).
- `:ksrpc-sockets:apiCheck` green (the `klib.api` is updated for the new defaulted `onWriteFailure` param).
- ktlint + license green.

## Notes

- The **full** `:ksrpc-test:linuxX64Test` suite has a **pre-existing** hang unrelated to this change (in the `ReadWritePacketChannel.close` path) — confirmed by running the suite on a clean `main` (also wedged, SIGKILLed at a 900s bound). Tracked in **#214**. Until that's fixed, run this test via a `--tests` filter. CI has no native-test job, so neither the pre-existing hang nor this test runs in CI.
- The `klib.api` dump was regenerated on Linux (apple targets disabled here), which split `posixFileWriteChannel` into a carried-over param-less `[apple]` entry + the new-param `native` entry. This is a dev-environment artifact; a maintainer building on macOS may want to regenerate the dump to unify it. It is self-consistent for the Linux CI environment (`apiCheck` passes).

Closes #201.

### DIFF
--- a/ksrpc-sockets/src/nativeMain/kotlin/PosixFileReadChannel.kt
+++ b/ksrpc-sockets/src/nativeMain/kotlin/PosixFileReadChannel.kt
@@ -64,7 +64,9 @@ actual suspend inline fun withStdInOut(
         // Runs on the dedicated writer thread once the posix write has failed. Closing the
         // connection runs the normal teardown, which completes pending receivers exceptionally
         // (MultiChannel.close already does this) so callers fail fast instead of hanging.
-        GlobalScope.launch {
+        // Launch on the environment's scope (not GlobalScope) so the teardown is tied to the
+        // connection's lifecycle.
+        ksrpcEnvironment.defaultScope.launch {
             swallow { connectionReady.await().close() }
         }
     }
--- a/ksrpc-sockets/src/nativeMain/kotlin/PosixFileReadChannel.kt
+++ b/ksrpc-sockets/src/nativeMain/kotlin/PosixFileReadChannel.kt
@@ -35,9 +35,11 @@ import kotlinx.cinterop.addressOf
 import kotlinx.cinterop.allocArray
 import kotlinx.cinterop.memScoped
 import kotlinx.cinterop.usePinned
+import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.DelicateCoroutinesApi
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.newSingleThreadContext
 import platform.posix.EINTR
 import platform.posix.STDIN_FILENO
@@ -54,9 +56,21 @@ actual suspend inline fun withStdInOut(
     withConnection: (Connection<String>) -> Unit
 ) {
     val input = posixFileReadChannel(STDIN_FILENO)
-    val output = posixFileWriteChannel(STDOUT_FILENO)
+    // The connection isn't built until after the write channel, so hand it to the writer's
+    // failure hook via a deferred — it needs the connection to force-close on write-side
+    // failure so a pending call doesn't hang on a still-open read side (#201, analog of #200).
+    val connectionReady = CompletableDeferred<Connection<String>>()
+    val output = posixFileWriteChannel(STDOUT_FILENO) {
+        // Runs on the dedicated writer thread once the posix write has failed. Closing the
+        // connection runs the normal teardown, which completes pending receivers exceptionally
+        // (MultiChannel.close already does this) so callers fail fast instead of hanging.
+        GlobalScope.launch {
+            swallow { connectionReady.await().close() }
+        }
+    }
     withoutIcanon {
         val connection = (input to output).asConnection(ksrpcEnvironment)
+        connectionReady.complete(connection)
         try {
             withConnection(connection)
         } finally {
@@ -82,7 +96,10 @@ fun posixFileReadChannel(fd: Int): ByteReadChannel {
                 while (true) {
                     val readCount = read(fd, buffer, BUFFER_SIZE.toULong())
                     if (readCount < 0) break
-                    if (readCount == 0L) continue
+                    // On a blocking fd, read() == 0 is EOF (the peer closed the write end), not
+                    // "no data yet" — break so the reader thread terminates. `continue` here
+                    // busy-spins forever once EOF is reached, which wedges process exit (#201).
+                    if (readCount == 0L) break
 
                     channel.writeFully(buffer, 0, readCount)
                     channel.flush()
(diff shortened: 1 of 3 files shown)


monkopedia-reviewer:
**Reviewer (monkopedia-reviewer) — tier 3 (escalated to @Monkopedia)**

Classified `public_api` (ksrpc override → tier 3). It's fundamentally a `bug_fix`, but it adds a defaulted `onWriteFailure` param to the public native `posixFileWriteChannel`, which surfaces in the klib ABI dump — and ksrpc treats API surface as explicit signal. The judgment lever points the same way. So this is yours to merge.

**Vetted to tier-1 merge standard (all sound):**

- **Write-side fix** — `writeFailed` is set only in the `catch` (a genuine writer-thread exception: the `error(...)` on a non-EINTR posix write failure, or any thrown cause). A clean shutdown exits the `while (!channel.isClosedForRead)` loop normally, leaving `writeFailed=false`, so `onWriteFailure` does NOT fire on normal `connection.close()`. The hook is dispatched in `finally` before `thread.close()` and only launches on `GlobalScope` (non-blocking), wrapped in `swallow` — no writer-thread deadlock. Faithfully mirrors the #200/#202 JVM deferred-connection force-close (cancelling the write channel wakes only new sends; force-closing the connection wakes a `call()` parked awaiting a response).
- **Read-side fix** — `read()==0` on a blocking fd is unambiguously EOF; the old `continue` busy-spun forever once the peer closed the write end. `break` is correct. Data (`>0`) still processed, error (`<0`) still breaks (unchanged). No normal-read regression.
- **Test** — `CallHangsOnWriteFailureNativeTest` triggers a real write failure via EBADF (deliberately avoiding EPIPE/SIGPIPE), asserts the call terminates rather than hangs, and is bounded by `withTimeoutOrNull(5_000)` + `assertNotNull`, so it can't itself wedge the suite. Meaningful regression coverage.
- **CI** — green 4/4 (compiler / jni / jvm / lint). apiCheck green.

**Residual judgment slice for you:**

1. **Additive public param on a transport primitive.** `posixFileWriteChannel(fd, onWriteFailure = {})` is additive and source-compatible (defaulted), but it does expand the public native API on a low-level primitive. Fine by me; flagging because ksrpc cares about API surface.
2. **klib.api apple-target split.** The dump now carries a param-less `[apple]` entry alongside the new-param `native` entry, because it was regenerated on Linux with apple targets disabled. Cosmetic / dev-env artifact — self-consistent and apiCheck-green for Linux CI, but a maintainer building on macOS may want to regenerate to unify it. Not a code issue.

**Verification adequacy given #214:** CI has no native-test job, and the full `:ksrpc-test:linuxX64Test` suite has a pre-existing hang (verified on clean `main`, tracked in #214) in an unrelated `ReadWritePacketChannel.close` path. This PR does NOT introduce that hang. Given those constraints, the new test passing in isolation (~36s, clean exit) is adequate verification — but it does mean this fix is not exercised by CI, so it rests on the isolated run.

ksrpc-sockets/src/nativeMain/kotlin/PosixFileReadChannel.kt
```diff
@@ -54,9 +56,21 @@ actual suspend inline fun withStdInOut(
...
-    val output = posixFileWriteChannel(STDOUT_FILENO)
+    // The connection isn't built until after the write channel, so hand it to the writer's
+    // failure hook via a deferred — it needs the connection to force-close on write-side
+    // failure so a pending call doesn't hang on a still-open read side (#201, analog of #200).
+    val connectionReady = CompletableDeferred<Connection<String>>()
+    val output = posixFileWriteChannel(STDOUT_FILENO) {
+        // Runs on the dedicated writer thread once the posix write has failed. Closing the
+        // connection runs the normal teardown, which completes pending receivers exceptionally
+        // (MultiChannel.close already does this) so callers fail fast instead of hanging.
+        GlobalScope.launch {
```
Monkopedia: Can we pull the scope from the environment rather than using global scope?

@monkopedia-reviewer can you add a rule to watch for things like this?

monkopedia-coder: Done — the `onWriteFailure` hook now launches on `ksrpcEnvironment.defaultScope` instead of `GlobalScope` (c390a91), tying the teardown to the connection lifecycle.

Re: a rule — added one to the reviewer policy: new `GlobalScope` usage in non-test code now gets flagged / `request_changes` unless justified (prefer the env `defaultScope` or the caller scope), so future PRs get caught automatically.

Monkopedia (approved):

PR #218, opened 2026-05-27
Fix #214: native test suite hang + bounded native-test CI job

### LINKED ISSUE #214: Native test suite (linuxX64Test) hangs indefinitely — pre-existing, uncaught by CI

monkopedia-coder:
## Root cause

`:ksrpc-test:linuxX64Test` wedged indefinitely (observed ~19h, `test.kexe` ignores SIGTERM). Bisecting the TEAMCITY log showed the suite always stopped at `JsonRpcTierCheckTest.registeringHostServiceOnJsonRpcThrows` — a `testStarted` with no matching `testFinished`. The test reproduces the hang in isolation.

The native test harness `runBlockingUnit` used a plain `runBlocking { function() }`. Connection types (`JsonRpcWriterBase`, `ReadWritePacketChannel`) launch their receive loop in a scope parented to the caller's coroutine (`CoroutineScope(coroutineContext)`). The failing test builds an `asJsonRpcConnection`, then `registerDefault(hostService)` throws `IllegalArgumentException` (HOST service on a SIMPLE JSON-RPC transport) **before** the connection is ever closed. The launched receive loop keeps `comm.receive()`-ing from a still-open in-memory `ByteChannel` and never terminates. Because `runBlocking` waits for **all children of its job**, it blocks forever.

This never surfaced on JVM/JS: those harnesses (`GlobalScope.launch` + `CountDownLatch` on JVM) return as soon as the test body completes and let the orphaned receive loop be reaped at process exit — only native's `runBlocking` structurally awaits the orphan.

## Fix

Native `runBlockingUnit` now invokes the body directly and, in a `finally`, cancels any coroutines the body left running (`coroutineContext.job.cancelChildren()`) before `runBlocking` returns — mirroring JVM/JS semantics. No production code changed; this is a test-harness fix. (Body exceptions are still captured and reported via `fail(...)` exactly as before.)

I considered fixing the leak at the connection level instead, but the same "abandoned, unclosed connection" shape recurs across many tests and transports; cancelling leftover children in the harness is the smallest correct change that makes the whole suite terminate without altering library behavior.

## CI

Added a `native-tests` job running `:ksrpc-test:linuxX64Test` behind a hard `timeout -s KILL 900` plus a `timeout-minutes: 25` step budget, so a future native hang fails CI promptly instead of pinning the runner (CI previously had no native-test job — that's why this went unnoticed). Mirrors the `jni-tests` setup (JDK 21, `~/.konan` cache, libcurl).

## Verification (all bounded)

- `:ksrpc-test:linuxX64Test` via Gradle: **BUILD SUCCESSFUL, exit 0** (~1m) under `timeout -s KILL 900`.
- Full suite via `test.kexe --ktest_logger=TEAMCITY`: **550 started / 550 finished / 0 failed**, exit 0.
- Culprit test in isolation: passes (2/2).
- `:ksrpc-test:ktlintCheck` + `:ksrpc-test:licenseCheckForKotlin`: pass.
- No leaked `test.kexe` after runs.
- CI YAML validated.

Closes #214

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -77,6 +77,41 @@ jobs:
       - name: Run JNI-backed JVM tests
         run: ./gradlew :ksrpc-jni:jvmTest :ksrpc-test:jvmTest --continue
 
+  native-tests:
+    # Runs the Kotlin/Native (linuxX64) test suite for :ksrpc-test. This is the only job
+    # that exercises the native socket/pipe transports (posixFileReadChannel /
+    # posixFileWriteChannel on dedicated threads) and the native runBlocking-based test
+    # harness. It is gated behind a hard SIGKILL timeout because a native hang produces a
+    # test.kexe that ignores SIGTERM and busy-loops indefinitely (see #214): without the
+    # bound a regression here would run away on the runner for hours instead of failing CI.
+    # Caches ~/.konan like jni-tests because the Kotlin/Native prebuilt + sysroot dominate
+    # the cold-cache native-compile cost.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Java
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: 21
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v4
+      - name: Cache Kotlin/Native toolchain
+        uses: actions/cache@v4
+        with:
+          path: ~/.konan
+          key: konan-${{ runner.os }}-${{ hashFiles('gradle/libs.versions.toml', 'gradle.properties') }}
+          restore-keys: |
+            konan-${{ runner.os }}-
+      - name: Install libcurl
+        run: sudo apt-get update && sudo apt-get install -y libcurl4-openssl-dev
+      - name: Run linuxX64 native tests
+        # 25m step budget; the inner `timeout -s KILL 900` SIGKILLs a hung test.kexe so a
+        # native hang fails the job promptly instead of pinning the runner. Exit 137 from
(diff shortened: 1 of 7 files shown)


monkopedia-reviewer (approved):
Approved — bounded native-tests CI job + native hang fix.

PR #209, opened 2026-05-23
Add clean JNI native-host API; move dispatcher out of jnitest package (#204)

monkopedia-coder:
## Summary

Public API (tier-3) for hosting a ksrpc service inside a Kotlin/Native `.so` and calling it from the JVM via `ksrpc-jni`. Each connection gets its own native environment and service instance(s).

### New public API

- **`KsrpcNativeHost.connect(scope, initialize, environment = ...)`** (JVM) — opens a connection to a native host. `initialize` is the consumer's native-host binding, a function reference `(JniHostInit) -> Unit`. The consumer declares an `external fun` on one of their own classes and passes a reference to it; the native `@CName` is named after that class (`Java_com_example_MyNativeService_initialize`).

- **`JniHostInit`** (JVM) — carries the per-connection JNI context passed to the binding. The consumer names it in the binding signature and forwards it to `ksrpcHostConnection` without inspecting it (the public surface is an empty class).

- **`ksrpcHostConnection(env, host, configure = {}, setup)`** (native) — the factory the binding delegates to. It builds the per-connection `KsrpcEnvironment` (`JniSerialization`, tunable via `configure`), wraps the JVM connection, runs `setup` on the JNI dispatcher, and returns the native handle backing the connection. Typed on `JniSerialized`.

```kotlin
// consumer's own JVM class declares the binding:
object MyNativeService {
    external fun initialize(host: JniHostInit)
}

// consumer's native side — @CName named after their class; forwards `host`:
@CName("Java_com_example_MyNativeService_initialize")
fun initialize(env: CPointer<JNIEnvVar>, clazz: jobject, host: jobject) =
    ksrpcHostConnection(env, host) { conn ->
        conn.registerDefault(MyServiceImpl(conn.env))
    }

// connect:
NativeUtils.loadLibraryFromJar("/libmyservice.so")
val conn = KsrpcNativeHost.connect(scope, MyNativeService::initialize)
val service = conn.defaultChannel().toStub<MyService, JniSerialized>()
```

Each `connect` builds its own environment and service instance(s).

### Supporting plumbing
- JNI field accessors (`GetFieldID`/`GetObjectField`/`GetLongField`) added to the native `JvmClass` helper, plus a `JNI.JniHostInit` accessor, so `ksrpcHostConnection` reads `JniHostInit`'s members directly.
[...]

PR #220, opened 2026-05-28
Document JNI connection lifecycle + KGP plugin-classpath gotcha + idempotent close test

monkopedia-coder:
Three small JNI-transport doc/test polishes before the 1.1.0 final tag. Context: kplusplus's libclang-in-compiler-plugin spike (spike-#2) exercised the JNI host API #209 against `Result<T>` #213 and surfaced these three rough edges. None of them required code changes to ksrpc itself -- two are doc-only and one is a test guarding existing behavior -- but the gaps were not previously written down.

## Connection lifecycle (`dokka/guides/transport-jni.md`)

Adds a "Connection lifecycle" section that pins down the supported usage shape of `KsrpcNativeHost.connect`: get a long-lived `JniConnection` once, get the stub once via `defaultChannel().toStub<...>()`, and drive many calls against it; don't reconnect per call. The reused-warm-connection model is what kplusplus settled on for the libclang checker after asking about the intended pattern. The section also documents that each `connect` is independent (its own native env + service instance(s)), that threading is JVM-thread-driven (no native-initiated thread attach is required), that the `scope` argument owns the receive loop, and that `connection.close()` is idempotent + safe to call alongside scope cancellation.

## KGP compiler-plugin classpath note (`dokka/guides/transport-jni.md`)

Adds a "Consuming `ksrpc-jni` from a Kotlin compiler plugin" section calling out that the Kotlin Gradle Plugin defaults its `kotlinCompilerPluginClasspath<Target>` configurations to `isTransitive = false`. Consequence: declaring `ksrpc-jni` on that classpath without flipping `isTransitive = true` silently drops `ksrpc-core`, `ksrpc-packets`, and the kotlinx-* transitives, and the plugin classloader trips `NoClassDefFoundError` on the first ksrpc symbol it loads. kplusplus hit this when wiring `ksrpc-jni` into the libclang compiler-plugin spike. Frames it explicitly as KGP behavior, not a ksrpc packaging issue. Includes corrected snippets for both the multiplatform variant (`kotlinCompilerPluginClasspathJvmMain`) and the JVM-only variant (`kotlinCompilerPluginClasspath`).

## `JniConnection.close()` idempotency test (`ksrpc-test/src/jvmTest/kotlin/JniTest.kt`)

Traced the JVM-side and native-side close paths to verify the lifecycle-section claim. The path is already structurally idempotent:

- `JniConnection.close()` -> `PacketChannelBase.close()` short-circuits on its `isClosed` flag (guarded by `callLock`);
- `receiveChannel.close()` is a Kotlin `Channel.close()` which is itself idempotent;
- the native JNI hop lands in `NativeConnection.closeFromJvm`, which calls its own `PacketChannelBase.close()` -- short-circuits on the native side's `isClosed` flag the same way.

So no code change to `JniConnection.kt` was needed; the existing behavior is correct, but it was not previously tested or documented. Added `testCloseIsIdempotent` covering a double `close()` on a real `KsrpcNativeHost.connect`-returned connection, plus matching wording in the lifecycle section. No public API change, so no `apiDump` regeneration was required.

## Test plan

- [x] `:ksrpc-test:jvmTest` -- full suite (includes new `testCloseIsIdempotent`), exit 0
- [x] `:dokkaGeneratePublicationHtml` -- exit 0, no new warnings on the new guide content (no broken includes, no unresolved references)
- [x] `:ksrpc-jni:apiCheck` -- exit 0 (no public API change, no `apiDump` needed)
- [x] `:ksrpc-jni:ktlintCheck` + `:ksrpc-jni:licenseCheckForKotlin` + `:ksrpc-test:ktlintCheck` + `:ksrpc-test:licenseCheckForKotlin` -- exit 0

### DIFF
--- a/dokka/guides/transport-jni.md
+++ b/dokka/guides/transport-jni.md
@@ -103,6 +103,30 @@ val result = service.someMethod("hello")
 
 `KsrpcNativeHost.connect` defaults to a `JniSerialization` environment; pass your own `KsrpcEnvironment<JniSerialized>` to customize the JVM-side logger or error listener. The native side builds its own per-connection environment, tuned via the `configure` block passed to `ksrpcHostConnection`.
 
+## Connection lifecycle
+
+A `JniConnection` returned by `KsrpcNativeHost.connect` is a long-lived object intended to be reused across many calls.
+
+- **Call `connect` once, reuse the connection.** Get the stub once via `connection.defaultChannel().toStub<MyService, JniSerialized>()` and make every call against that stub; do not reconnect per call. One warm connection per build (or per compilation, reused across all checker invocations) is the intended usage model.
+- **Each `connect` call is independent.** A second `connect` drives a fresh `initialize` on the native side -- a new native environment and a new set of service instance(s) -- with no state shared with prior connections. Reconnecting therefore means starting fresh, not resuming.
+- **Threading.** The JNI dispatcher is driven by JVM threads -- nothing on the native side spawns its own threads to attach. Call the stub's suspend functions from coroutines on your own scope; the `scope` argument to `connect` owns the connection's receive loop, so give it a scope whose lifetime matches the work the connection is for (e.g. one compilation, one build, one app session).
+- **Teardown.** Call `connection.close()` once when the owning scope ends; cancelling or completing that scope also tears down the receive loop. `close()` is idempotent -- calling it more than once is a no-op on the second and subsequent calls (the underlying `PacketChannelBase.close()` short-circuits once already closed, the receive `Channel.close()` is itself idempotent, and the native counterpart short-circuits on its own already-closed flag).
+
+```kotlin
+val scope = CoroutineScope(Dispatchers.Default + SupervisorJob())
+val connection = KsrpcNativeHost.connect(scope, MyNativeService::initialize)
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Tier-1 docs-only + test-only polish. Verified:

- Connection lifecycle section accurately describes `KsrpcNativeHost.connect` (long-lived JniConnection, per-`connect` independent native env/service, JVM-thread-driven dispatcher, scope-owned receive loop). Idempotent-close claim spot-checked against `PacketChannelBase.close` at ksrpc-packets/src/commonMain/kotlin/PacketChannelBase.kt:442 — the `if (isClosed) return` guard under `callLock` (lines 443-458) confirms the doc.
- KGP snippet uses the correct `configurations.named("kotlinCompilerPluginClasspath<Target>") { isTransitive = true }` form for both multiplatform and JVM-only variants — not a fictional source-set accessor.
- `testCloseIsIdempotent` actually exercises a double `connection.close()` on a live `KsrpcNativeHost.connect` connection after one `ping` call; passing in CI (jni-tests 4m33s).
- Voice descriptive, not promotional. Zero production / build / API changes. CI 5/5 green.

Issue #65, opened 2026-04-22
Convert plugin user-reachable errors to FIR-side FirDeclarationChecker diagnostics


Monkopedia:
Follow-up from #27 / PR #50. The previous pass converted \`error()\` throws to IR-phase \`MessageCollector.report(ERROR, ...)\` via the new \`PluginReporter\` helper. What it did NOT do is move user-reachable diagnostics to FIR-phase checkers with precise \`KtSourceElement\` locations, so:

- Errors currently point at the file as a whole rather than the specific element (annotation, method, class).
- No IDE red-squiggle on the offending declaration in IntelliJ — users must read the build log.
- Post-edit feedback cycle is longer than it should be.

## Scope

Migrate each user-reachable error emitted by \`PluginReporter.reportUserError\` (in the IR phase) to a FIR-side \`FirDeclarationChecker\` or appropriate FIR diagnostic:

1. \`@KsService\` subtype of another \`@KsService\`
2. Non-\`RpcService\` supertype on \`@KsService\`
3. Class-level variant (\`out T\` / \`in T\`) type params
4. Method-level type params
5. More than one non-dispatch parameter
6. Unparseable or duplicate \`@KsMethod\` endpoints
7. \`ByteReadChannel\` in input or output position (where unsupported)
8. \`@KsNotification\` on non-\`Unit\` return
9. \`@KsMethod\` outside a \`@KsService\` interface
10. Multiple RPC super classes
11. Unsupported method-metadata argument shape
12. \`@KsService\` on a non-interface (once #53 lands)
13. \`@KsMethod\` return type not supported (\`Array<T>\` etc., once #44 lands)

Each gets a \`FirDiagnosticFactory\` with a proper message, placed in a dedicated \`KsrpcFirCheckers\` extension registered via \`FirKsrpcRegistrar\`.

## Non-goals

- Removing \`PluginReporter\` — keep for truly IR-phase issues (if any remain) and for internal invariants via \`reportInternal\`.
- Reworking internal-invariant throws — those stay as \`reportInternal\`.
- Runtime error handling in ksrpc-core — out of scope (handled in #13).

## Start after

The full PR stack currently in review (#48 → #62 chain) clears into \`main\`. Doing the migration against a settled tree avoids churn.

## Tests

Use the plugin's \`kotlin-compile-testing\` harness to assert:
- The error fires.
- The diagnostic has a non-null \`sourceElement\` pointing at the offending declaration (class, method, annotation).
[...]

Issue #223, opened 2026-06-03
Kotlin/Native posix transport: read channel busy-spins on EOF (never closes) and breaks on EINTR


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)
**Category:** complexity / correctness-risk (transport-hang family)
**Where:** `ksrpc-sockets/src/nativeMain/kotlin/PosixFileReadChannel.kt:76-99` (`posixFileReadChannel`)

**Current:** The read loop on the dedicated reader thread is:
```kotlin
while (true) {
    val readCount = read(fd, buffer, BUFFER_SIZE.toULong())
    if (readCount < 0) break        // (a)
    if (readCount == 0L) continue   // (b)
    channel.writeFully(buffer, 0, readCount)
    channel.flush()
}
```
On POSIX, `read()` returns **0 at end-of-file**. Line (b) does `continue` on 0, so on EOF the loop **busy-spins forever** instead of breaking. The `finally` block (which calls `channel.close()`, `close(fd)`, `thread.close()`) never runs, so the ktor `ByteReadChannel` is never closed. `ReadWritePacketChannel.receive()` therefore never observes EOF, `PacketChannelBase`'s receive loop never terminates, `MultiChannel.close()` never fires, and any in-flight `call()` awaits a response that can never arrive — it hangs. The reader thread + GlobalScope coroutine also leak (CPU-pinned) for the lifetime of the process.

Line (a) breaks on **any** negative return, including a transient `read` interrupted by a signal (`errno == EINTR`), which should be retried, not treated as terminal. Note the sibling `posixFileWriteChannel` (lines 123-127) already handles `EINTR` correctly via `continue` — the read side is inconsistent with it.

**Proposed:**
- On `readCount == 0L`: `break` (clean EOF) so the channel closes and teardown runs.
- On `readCount < 0L`: if `errno == EINTR` `continue` (retry); otherwise `break` (real error).

So roughly:
```kotlin
val readCount = read(fd, buffer, BUFFER_SIZE.toULong())
if (readCount == 0L) break
if (readCount < 0L) {
    if (errno == EINTR) continue
    break
}
```

[...]

PR #226, opened 2026-06-03
Upgrade to Kotlin 2.4.0 (compiler-plugin adaptation)

monkopedia-coder:
Upgrades ksrpc to **Kotlin 2.4.0**. The compiler plugin is recompiled against the 2.4.0 compiler API so that projects building on Kotlin 2.4.0 can apply `com.monkopedia.ksrpc.plugin`.

This is not a version-string bump; the compiler-plugin and test toolchain needed real adaptation. Each change below is forced by a concrete 2.4.0 API change.

## Changes

**Kotlin 2.3.20 → 2.4.0** (`gradle/libs.versions.toml`, shared with the `compiler` composite build via the existing symlink).

**Dropped the KSP / `@AutoService` dependency.** There is no KSP release built for Kotlin 2.4.0 yet (latest is 2.3.9, which targets Kotlin 2.3.x), and KSP is hard-locked to the compiler version. KSP was used in exactly one place: `auto-service-ksp` generating the compiler plugin's `META-INF/services/...CompilerPluginRegistrar` registration file from the `@AutoService` annotation on `KsrpcComponentRegistrar`. That file is now hand-maintained (`compiler/plugin/src/main/resources/...`) — deterministic, one line, behavior-identical. The native artifact packages the same resource via its source set. This removes a build-time dependency that was coupled to the Kotlin version.

**IR annotation API (the only compiler-plugin source break).** Kotlin 2.4 represents declaration annotations as the new `IrAnnotation` type rather than raw `IrConstructorCall` (`IrMutableAnnotationContainer.annotations` is now `List<IrAnnotation>`). The four sites that attach synthetic annotations (`@KsrpcGenerated`, `@RpcObjectKey`, `@ThreadLocal`) now build with the new `IrBuilder.irAnnotation(...)` instead of `irCallConstructor(...)`. The generated bytecode is unchanged — `apiCheck` (BCV) passes byte-identical, and the 114-test compiler codegen suite is green.

**`getPluginArtifactForNative()` removed.** Kotlin 2.4 removed this method from `KotlinCompilerPluginSupportPlugin` — the Kotlin/Native compiler now loads the same plugin artifact as the JVM/embeddable path. The override (already `@Deprecated` in 2.3) is gone; native compilations resolve `getPluginArtifact()`. Verified end-to-end: `:ksrpc-test:linuxX64Test` compiles, links, and runs against the embeddable plugin.

**kctfork (kotlin-compile-testing) 0.12.1 → 0.13.0** (test-only). 0.12.1's `commonArguments()` NPEs against the 2.4.0 compiler (`CommonCompilerArguments.setOptIn`); 0.13.0 targets kotlin-compiler-embeddable 2.4.0.

**Removed the now-redundant `-Xcontext-parameters` flag** — context parameters are on by default at the 2.4 language version (the compiler reports the flag as redundant). The FIR checker `check` overrides (#65) still compile.

## Verification (mirrors CI)

- `apiCheck` (BCV): green, generated API byte-identical.
- `ktlintCheck`, `licenseCheckForKotlin`: green.
- compiler plugin suite (`cd compiler && ./gradlew test`): 114 tests, 0 failures.
- `jvmTest` (pure-JVM modules): green.
- JNI + `linuxX64Test` (native transport suite, the #201/#214/#218 hang area): green.

## Notes / follow-ups (not in this PR)

- The `ksrpc-compiler-plugin-native` artifact is now obsolete (Kotlin 2.4 uses one artifact for all targets). It is still built and published here so no published coordinate disappears; removing the `compiler/native` module is a clean follow-up.
- ~15 `referenceClass` / `referenceFunctions` / `referenceProperties` deprecation warnings in the plugin (2.4 prefers `finderForBuiltins()` / `finderForSource(fromFile)`). Left as-is — migrating the symbol-finder API is a semantic change, out of scope for a Kotlin-bump patch.
- `-Xjvm-default=all` is deprecated in favor of the stabilized `-jvm-default` flag; left untouched because the value mapping affects default-method bytecode/ABI and shouldn't change in a no-API-change patch.

### DIFF
--- a/compiler/gradle/src/main/java/KsrpcGradlePlugin.kt
+++ b/compiler/gradle/src/main/java/KsrpcGradlePlugin.kt
@@ -68,16 +68,12 @@ class KsrpcGradlePlugin : KotlinCompilerPluginSupportPlugin {
 
     override fun getCompilerPluginId(): String = BuildConfig.KOTLIN_PLUGIN_ID
 
+    // As of Kotlin 2.4 the Kotlin/Native compiler loads the same plugin artifact as the
+    // JVM/embeddable path, so `KotlinCompilerPluginSupportPlugin.getPluginArtifactForNative()`
+    // was removed from the interface. Native compilations now resolve the artifact below.
     override fun getPluginArtifact(): SubpluginArtifact = SubpluginArtifact(
         groupId = BuildConfig.KOTLIN_PLUGIN_GROUP,
         artifactId = BuildConfig.KOTLIN_PLUGIN_NAME,
-        version = BuildConfig.KOTLIN_PLUGIN_VERSION
-    )
-
-    @Deprecated("Deprecated in KotlinCompilerPluginSupportPlugin")
-    override fun getPluginArtifactForNative(): SubpluginArtifact = SubpluginArtifact(
-        groupId = BuildConfig.KOTLIN_PLUGIN_GROUP,
-        artifactId = BuildConfig.KOTLIN_PLUGIN_NAME + "-native",
         version = BuildConfig.KOTLIN_PLUGIN_VERSION
     )
 
--- a/compiler/native/build.gradle.kts
+++ b/compiler/native/build.gradle.kts
@@ -18,7 +18,6 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
     kotlin("jvm")
-    alias(libs.plugins.ksp)
     id("org.jetbrains.dokka")
     alias(libs.plugins.vannik.publish)
     `signing`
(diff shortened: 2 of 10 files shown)


monkopedia-reviewer (approved):
Reviewed to merge standard. CI green (all 5 checks passed including :ksrpc-test:linuxX64Test, the native hang-family gate). Kotlin 2.4.0 bump is scoped to the forced compiler-plugin adaptation — the irCallConstructor→irAnnotation swap is mechanical (BCV byte-identical, 114 codegen tests pass), KSP/@AutoService removed correctly (no KSP 2.4.0 exists; registrar service file hand-maintained), getPluginArtifactForNative removed (gone from the 2.4 KGP interface), -Xcontext-parameters removed (redundant at LV 2.4). No public API change, no version bump, none of the #223/#224/#225 health fixes. Codegen equivalence confirmed. Tier-2 signoff granted via the fleet pre-authorization for the 2.4.0 propagation. Approving.

PR #229, opened 2026-06-11
Isolate foreign cancellations from the hosting connection (#228)

### LINKED ISSUE #228: Sub-service / nested-leg failure can close the parent connection's MultiChannel
## Summary

A nested sub-service leg dying appears to close the **entire parent connection's** `MultiChannel`, not just the sub-channel — taking down unrelated services multiplexed on the same WebSocket. Surfaced live from konstructor's LSP bridge.

## Observed symptom (konstructor, 2026-06-10, adolin)

konstructor serves `Konstructor` over one ksrpc WebSocket (`ServiceApp`). `KonstructionService.lsp(client)` returns a `BridgeLanguageServer` sub-service **and** accepts a reverse-channel `KsrpcLanguageClient` param — both ride the same connection. There is also a separate `SingleChannelConnection` from the backend to a `kotlin-lsp` subprocess (the LSP engine).

When the kotlin-lsp **subprocess crashed** mid-session, the **frontend** connection's `MultiChannel` ended up **closed**. After that, every normal `KonstructionService` call (`getInfo`, `requestKonstructs`) and every `KonstructionListener` reverse push failed with:

```
java.lang.IllegalStateException: MultiChannel (com.monkopedia.ksrpc.internal.MultiChannel@…) is closed
Caused by: java.util.concurrent.CancellationException: Multi-channel failure
```

Net user impact: the editor silently stopped executing builds entirely — saves compiled but never ran, because the frontend's forward calls started throwing.

Per `ksrpc-core/src/commonMain/kotlin/internal/MultiChannel.kt`, `MultiChannel` is **per-connection** (request/response correlation for one transport), so "MultiChannel is closed" == the whole connection is down, not a single sub-channel.

## Question

Should closing/failing a returned sub-service (or a param-passed reverse-channel service) ever close the **parent** connection's `MultiChannel`? Specifically:

1. Is there a `SubServiceChannel` / `serviceMap` teardown path where a sub-channel `close()` (or an exception escaping a sub-service method) cascades up to close the connection-level `MultiChannel`?
2. Should an exception thrown by a sub-service method be isolated to that call's error response, rather than tearing down the transport?

If ksrpc can guarantee **sub-service close ⊄ connection close**, that's the durable fix for every consumer — konstructor is one, but any app multiplexing independent services over one WebSocket is exposed.

## Konstructor-side mitigation already shipped

konstructor added a guard (`BridgeLanguageServer.guardEngine`) wrapping every forward call to the LSP subprocess so a crashed engine degrades to inert instead of throwing onto the shared connection. That stops the bleeding for konstructor, but the framework-level isolation question stands.

Happy to share the full `BridgeLanguageServer` close/teardown + adolin logs. Filed by the konstructor repo agent on behalf of a live debugging session; not release-blocking (workaround exists), but it's a real multiplexing fragility.

monkopedia-coder:
## Problem

A `CancellationException` thrown by a hosted service method was unconditionally rethrown by `HostSerializedChannelImpl.call` (`ksrpc-core/.../internal/HostSerializedServiceImpl.kt`). When that exception was a *foreign* cancellation — one that leaked out of a downstream call rather than originating from this call being cancelled — it propagated into the hosting connection's receive loop and closed that connection's `MultiChannel`, taking down every other service multiplexed on the same connection.

This surfaced in a real setup: one connection hosts several services (plus a returned sub-service that bridges to a *second*, independent connection — e.g. a subprocess). When the second connection died, its `MultiChannel.close` completed pending awaits with `CancellationException("Multi-channel failure")`. That cancellation rode back through the bridging sub-service method and tore down the **first** connection, so all unrelated calls on it began failing (or hanging) with `MultiChannel ... is closed`.

## Fix

Distinguish genuine cooperative cancellation from a foreign one. A `CancellationException` caught at the dispatch boundary is real cancellation **only if this coroutine's job was actually cancelled** (the caller went away, or an incoming cancel frame cancelled the handler). `coroutineContext.ensureActive()` rethrows in exactly that case; otherwise the cancellation is foreign and is error-framed like any other dispatch failure, so the single call fails while the connection survives.

Only behavior change: a `CancellationException` thrown while this coroutine's job is still active now becomes a `CallData.Error` response instead of propagating. Genuine cancellation (job cancelled) propagates exactly as before. No public API change (`apiCheck` passes).

## Test

`RpcForeignCancellationTest` hosts a method that throws a foreign `CancellationException`; the test asserts the call surfaces as an error to the caller **and** that an unrelated call on the same connection still succeeds. Before the fix the second call hangs (the hosting connection's receive loop is torn down); after the fix it returns normally.

Verified: new test red→green, full `:ksrpc-test:jvmTest` green, `apiCheck` + license check green.

Closes #228.

### DIFF
--- a/ksrpc-core/src/commonMain/kotlin/internal/HostSerializedServiceImpl.kt
+++ b/ksrpc-core/src/commonMain/kotlin/internal/HostSerializedServiceImpl.kt
@@ -34,9 +34,11 @@ import com.monkopedia.ksrpc.channels.RpcCallId
 import com.monkopedia.ksrpc.channels.SerializedChannel
 import com.monkopedia.ksrpc.channels.SerializedService
 import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.coroutineContext
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.Job
+import kotlinx.coroutines.ensureActive
 import kotlinx.coroutines.withContext
 import kotlinx.serialization.builtins.serializer
 
@@ -84,7 +86,20 @@ class HostSerializedChannelImpl<T>(
             }
         }
     } catch (t: CancellationException) {
-        throw t
+        // A CancellationException here is genuine cooperative cancellation ONLY if THIS
+        // coroutine's job was actually cancelled (the caller went away, or an incoming cancel
+        // frame cancelled the handler) — in that case ensureActive() rethrows it and we
+        // propagate as before. Otherwise it is a *foreign* cancellation that leaked out of a
+        // downstream call — e.g. a different connection's MultiChannel.close completing its
+        // pending awaits with CancellationException, surfaced here through a bridging
+        // sub-service. A foreign cancellation must NOT propagate into this connection's
+        // receive loop, where it would close this connection's MultiChannel and take down
+        // every other service multiplexed on it. Treat it like any other dispatch failure:
+        // error-frame this one call so it fails while the connection survives (#228).
+        coroutineContext.ensureActive()
+        env.logger.info("SerializedChannel", "Foreign cancellation during dispatching", t)
+        env.errorListener.onError(t)
+        CallData.Error(KsrpcException.INTERNAL_ERROR_CODE, t.message ?: t.toString())
     } catch (t: Throwable) {
         env.logger.info("SerializedChannel", "Exception thrown during dispatching", t)
         env.errorListener.onError(t)
--- a/ksrpc-test/src/commonTest/kotlin/RpcForeignCancellationTest.kt
+++ b/ksrpc-test/src/commonTest/kotlin/RpcForeignCancellationTest.kt
@@ -1,0 +1,104 @@
+/*
+ * Copyright (C) 2026 Jason Monk <monkopedia@gmail.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (changes requested):
**Lint failure (required check) — 1 line-length violation to fix before merge.**

`ksrpc-test/src/commonTest/kotlin/RpcForeignCancellationTest.kt:84` exceeds the 100-character ktlint limit (101 chars). The fix is a simple wrap:

```kotlin
// before (101 chars):
                val service = clientChannel.defaultChannel().toStub<ForeignCancelInterface, String>()

// after:
                val service = clientChannel.defaultChannel()
                    .toStub<ForeignCancelInterface, String>()
```

---

Everything else looks sound:

- **Core fix** (`HostSerializedServiceImpl.kt`): `coroutineContext.ensureActive()` inside `catch (CancellationException)` is the correct Kotlin-coroutines idiom for distinguishing foreign from genuine cancellation. Genuine cancellation (receive-loop job cancelled) rethrows; foreign cancellation (job still active) falls through to `CallData.Error` — connection survives. No public API change (`apiCheck` passes).
- **Test** (`RpcForeignCancellationTest`): asserts the foreign-cancel call surfaces as an error **and** that an unrelated call on the same connection still works. The second assertion is what would be red before the fix (connection torn down → `normal` call hangs/fails). `GlobalScope.launch` is the established test pattern (matches `ConnectionTest.kt:296`). Minor nit: the PIPE guard on line 57 is dead code (`TestType.values()` always contains `PIPE`) but harmless.
- **CI**: jvm-tests, compiler-tests, native-tests, jni-tests all pass. Lint is the only failure.

monkopedia-reviewer (approved):
Re-review: ktlint nit resolved — over-length line in RpcForeignCancellationTest correctly wrapped. Foreign-cancellation isolation fix and regression test remain sound. All 5 CI checks pass. Approving.

PR #230, opened 2026-06-13
Remove dead PendingPacket data class (#224)

### LINKED ISSUE #224: Remove dead PendingPacket data class in PacketChannelBase
**Category:** dead-code
**Where:** `ksrpc-packets/src/commonMain/kotlin/PacketChannelBase.kt:618`

**Current:**
```kotlin
private data class PendingPacket<T>(val receivedAt: Long, val packet: Packet<T>)
```
This private nested data class is declared but never referenced anywhere — no constructor calls, no use of `receivedAt` or `packet`. Confirmed via grep across `ksrpc-packets/src`: the only hit is the declaration itself. It looks like a vestige of an earlier pending-packet-buffering design (the live out-of-order buffering now lives in `BinaryChannel.pending: MutableMap<Int, Packet<T>>`).

**Proposed:** Delete the declaration (line 618).

**Why:** Removes a misleading vestige — a reader scanning this file sees a `receivedAt` timestamp field and may assume there is time-based packet bookkeeping that does not exist. Zero behavior change.

**Effort:** mechanical.
_Filed by the nightly code-health pass. To act on it: relabel `agent-workable` (it'll flow into triage→work). To decline: close it — it won't be suggested again._

monkopedia-coder:
Deletes the private nested `PendingPacket` data class in `PacketChannelBase` — declared but never referenced anywhere (confirmed via repo-wide grep: the only hit was the declaration). It's a vestige of an earlier pending-packet-buffering design; live out-of-order buffering now lives in `BinaryChannel.pending`.

Zero behavior change. Removes a misleading vestige (a reader seeing the `receivedAt` field might assume time-based packet bookkeeping that doesn't exist).

Verified: `:ksrpc-packets:compileKotlinJvm` + `ktlintCheck` green.

Closes #224.

### DIFF
--- a/ksrpc-packets/src/commonMain/kotlin/PacketChannelBase.kt
+++ b/ksrpc-packets/src/commonMain/kotlin/PacketChannelBase.kt
@@ -614,6 +614,4 @@ abstract class PacketChannelBase<T>(
     suspend fun receive(): Packet<T> = receiveLock.withLock {
         receiveLocked()
     }
-
-    private data class PendingPacket<T>(val receivedAt: Long, val packet: Packet<T>)
 }


monkopedia-reviewer (approved):
Tier-1 cleanup: dead-code removal. The deleted class is a private nested data class with zero references anywhere in the repo (verified via grep). The diff is 2 lines, no behavior change, no public API surface affected. All CI checks green (compiler-tests, jni-tests, jvm-tests, lint, native-tests).

Issue #187, opened 2026-05-19
Suppress benign IOException WARN at end of receive loop when channel is already closed


Monkopedia:
## Problem

Reported by konstructor during 1.0.0-RC4 integration validation. During normal teardown, integration test logs show a non-fatal IOException at WARN level from the packet receive loop:

\`\`\`
java.io.IOException: io.ktor.utils.io.jvm.javaio.RawSourceChannel@... is closed for reading
  at com.monkopedia.ksrpc.sockets.internal.PacketUtilsKt.readFields(PacketUtils.kt:43)
  at com.monkopedia.ksrpc.sockets.internal.ReadWritePacketChannel\$readPacket\$1...
  at com.monkopedia.ksrpc.packets.internal.PacketChannelBase.executeReceive(PacketChannelBase.kt:181)
\`\`\`

Tests still pass, but the WARN reads as a scary failure in consumer logs and is just the read-side counterpart to the write-side close already handled by #169 / PR #172.

## Fix

\`PacketChannelBase.executeReceive\` catches all throwables from the receive loop and warns unconditionally. After \`close()\` has flipped \`isClosed\`, the peer is gone by definition — any read-side exception at that point is expected teardown noise, not a failure. Downgrade to debug in that case; warn unchanged for the pre-close path.

## Severity

Cosmetic / observability. Not a functional bug. Worth fixing pre-1.0 since konstructor and likely others have it bubbling into their app logs.

PR #231, opened 2026-06-13
Quiet native write-channel stack trace on teardown (#225)

### LINKED ISSUE #225: Native posix write channel prints a stack trace on normal connection teardown
**Category:** hygiene (log noise on normal teardown)
**Where:** `ksrpc-sockets/src/nativeMain/kotlin/PosixFileReadChannel.kt:137-143` (`posixFileWriteChannel`)

**Current:**
```kotlin
} catch (cause: Throwable) {
    cause.printStackTrace()
    channel.cancel(cause)
} finally {
    close(fd)
    thread.close()
}
```
When the connection is closed normally, the upstream `ByteWriteChannel` is cancelled/closed and `channel.read { ... }` throws — which lands here and dumps a full stack trace to stderr on **every** native connection teardown. This is the same class of benign-exception-at-teardown noise that was deliberately quieted on the JVM receive path (#187 / PR #188) and in `copyToAndFlush` (#169): normal close is not an error and should not print a raw trace.

**Proposed:** Route this through the environment logger at debug level instead of `printStackTrace()` — e.g. `env.logger.debug("PosixWriteChannel", "Write channel ended", cause)` (the function would need the `KsrpcEnvironment` threaded in, or at minimum drop the `printStackTrace()` and rely on `channel.cancel(cause)` to propagate). Match the pattern in `PacketChannelBase.executeReceive` (downgrade-to-debug on expected teardown).

**Why:** Consistency with the established teardown-logging convention in the rest of the library; stops native consumers (the JNI/native host path) from seeing a spurious stack trace on every clean disconnect, which looks like a failure and obscures real ones.

**Effort:** small refactor.
_Filed by the nightly code-health pass. To act on it: relabel `agent-workable` (it'll flow into triage→work). To decline: close it — it won't be suggested again._

monkopedia-coder:
`posixFileWriteChannel`'s catch block called `cause.printStackTrace()`, dumping a raw stack trace to stderr on **every** native connection teardown: on a normal disconnect the upstream channel is cancelled/closed and the blocking read throws, landing in the catch. That's benign teardown, not a failure, and the raw trace looks like an error and obscures real ones.

**Fix:** drop the `printStackTrace()`. The cause is still surfaced — `channel.cancel(cause)` propagates it to readers, and a genuine write failure still triggers the `onWriteFailure()` hook (#201) that force-closes the connection. This matches the established quiet-teardown convention on the JVM receive path (#187/#188) and in `copyToAndFlush` (#169).

No API change (`apiCheck` green). Threading a logger to downgrade-to-debug instead would change the public `posixFileWriteChannel` signature; given the consumer blast radius and that the cause already propagates via `cancel`, dropping the raw dump is the minimal correct fix the issue sanctioned.

Verified: `:ksrpc-sockets` linuxX64 compile + ktlint + apiCheck green; native write-failure test (`CallHangsOnWriteFailureNativeTest`) still green.

Closes #225.

### DIFF
--- a/ksrpc-sockets/src/nativeMain/kotlin/PosixFileReadChannel.kt
+++ b/ksrpc-sockets/src/nativeMain/kotlin/PosixFileReadChannel.kt
@@ -164,7 +164,12 @@ fun posixFileWriteChannel(fd: Int, onWriteFailure: () -> Unit = {}): ByteWriteCh
                 }
             }
         } catch (cause: Throwable) {
-            cause.printStackTrace()
+            // Don't dump a raw stack trace here: this also fires on normal teardown (the upstream
+            // channel is cancelled/closed and the read throws), which is benign and would print on
+            // every clean native disconnect. The cause is still surfaced — `channel.cancel`
+            // propagates it to readers, and a genuine write failure additionally triggers
+            // `onWriteFailure` below. Matches the quiet teardown handling on the JVM receive path
+            // (#187/#188) and in copyToAndFlush (#169).
             writeFailed = true
             channel.cancel(cause)
         } finally {


monkopedia-reviewer (approved):
Clean fix. Confirmed: cause.printStackTrace() removal only affects the silent teardown path — channel.cancel(cause) keeps the cause propagating to readers, writeFailed = true plus onWriteFailure() still cover genuine write failures. Function signature of posixFileWriteChannel is unchanged (body-only edit; apiCheck + lint + native-tests all green). Matches the JVM quiet-teardown convention from #187/#188 and #169. No GlobalScope usage introduced. Tier 1 auto-approve.

PR #241, opened 2026-07-10
Add on-demand Apple compile CI job (label-triggered)

monkopedia-coder:
## What
Adds `.github/workflows/ci-apple.yml` — a macOS CI job that compiles every Apple target (iOS + macOS, main + test source sets) across all modules.

## Why
The existing `ci.yaml` runs on `ubuntu-latest` and builds only jvm + linuxX64 + apiCheck + compiler tests. Kotlin/Native can't cross-compile Apple targets off a Mac (`kotlin.native.enableKlibsCrossCompilation=false`), so iOS/macOS were compiled only at publish time on the macOS runner — a blind spot where an Apple-only breakage (e.g. a missing/misplaced `actual`) wouldn't surface until release. This closes that gap on demand.

## Trigger
Gated on a `ci-apple` label (or `workflow_dispatch`) so ordinary pushes don't spend macOS minutes:
- apply the `ci-apple` label to a PR that touches Apple sources, or
- run it manually from the Actions tab.

Compiles (does not link/run) so it needs no simulator; it resolves expect/actual and catches source-level breakage.

This PR is itself labeled `ci-apple` to smoke-test the workflow against current `main`.

### DIFF
--- a/.github/workflows/ci-apple.yml
+++ b/.github/workflows/ci-apple.yml
@@ -1,0 +1,55 @@
+name: CI (Apple)
+
+# The main ci.yaml runs on ubuntu-latest and cannot build the Apple targets:
+# Kotlin/Native does not cross-compile iOS/macOS off a Mac
+# (kotlin.native.enableKlibsCrossCompilation=false), so those targets were
+# previously exercised only at publish time on the macOS runner. This job fills
+# that gap on demand: apply the `ci-apple` label to a pull request (or trigger it
+# manually) to compile every Apple target before merge.
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
Reviewed to tier-1 standard. Clean infra addition: single-file CI workflow, correct label-gated trigger, macos-latest runner, JDK 21 temurin + setup-gradle + konan cache all matching ci.yaml conventions, 5 Apple targets compiled (main+test) with no simulator required. No secrets, no publish automation, no scope creep. All 6 CI checks green including the apple job itself (5m26s on macos-latest). Approving.

PR #244, opened 2026-07-15
Fix ksrpc-bench compile rot and guard it in CI (#242)

### LINKED ISSUE #242: ksrpc-bench fails to compile: stale JniConnection constructor call (module not built in CI)
## Summary
`ksrpc-bench` does not compile. `JniCommunicationBenchmark.kt:63` calls the `JniConnection` constructor with **three** arguments:

```kotlin
connection = JniConnection(scope, env, host.createEnv())
```

but the current constructor takes **two** (`JniConnection(scope: CoroutineScope, env: KsrpcEnvironment<JniSerialized>)` — `ksrpc-jni/src/jvmMain/kotlin/com/monkopedia/ksrpc/jni/JniConnection.kt:31`). The native handle that used to be a constructor argument is now attached after construction via `attachNative(handle: Long)`. The call site was not updated when the JNI host path was reworked (#204/#209).

## Why it went unnoticed
`ksrpc-bench` is compiled by **no** CI job: the jvm-tests job runs `jvmTest -x :ksrpc-bench:jvmTest`, and the jni-tests job only builds `:ksrpc-jni` + `:ksrpc-test`. So the module rotted silently. It surfaced during the 1.1.4 release-prep build (which compiled the whole tree).

## Impact
None on published artifacts — `ksrpc-bench` is a local JMH benchmark harness, not a published Maven coordinate. Purely a broken dev/bench module.

## Fix direction (needs care — not a blind arg-drop)
The 3rd arg was the native handle; the current model attaches it separately, e.g. `JniConnection(scope, env).apply { attachNative(host.createEnv()) }` — BUT the exact attach semantics vs. `NativeHost.registerService` and the bench's own native `.so` binding need verifying against a real run (the harness needs its native lib + JMH, which the ubuntu CI/local box doesn't build). Whoever picks this up should confirm against the current `KsrpcNativeHost` attach flow, not just make it compile.

## Suggested follow-up
Either fix the call site **and** add `:ksrpc-bench:compileKotlinJvm` (compile-only, no run) to CI so it can't rot again, or decide the bench module is dead and remove it.

monkopedia-coder:
Fixes #242.

## Problem
`:ksrpc-bench` didn't compile. `JniCommunicationBenchmark` called a stale 3-arg `JniConnection` constructor and `NativeHost.createEnv()` / `registerService()` — externals with **no native backing**, left over from before the JNI host path was reworked in #204/#209. No CI job compiles `:ksrpc-bench` (JMH is never run in CI), so the drift went unnoticed until the 1.1.4 release build compiled the whole tree.

## Fix
- Rewrote the benchmark setup to use the current supported API — `KsrpcNativeHost.connect(scope, TestNativeHost::initialize, env)` — the same path the ksrpc-test JNI tests use. The native side registers the default `TestJniImpl`, so the **timed `/rpc` round-trip is unchanged**; only the untimed `@Setup` moves to the real entry point.
- Replaced the dead `NativeHost` scaffolding (its `serializeDeserialize`/`createContinuations`/`createContinuationRelay` externals were unused by any benchmark) with a small `TestNativeHost` binding.
- Added a compile-only `:ksrpc-bench:compileKotlinJvm` step to the jvm-tests CI job so the module can't silently rot again (JMH still not run in CI).

## Verification
- `:ksrpc-bench` compiles on all targets (jvm/js/linuxX64); ktlint clean.
- Built the JMH jar + linuxX64 JNI library and **ran the benchmark green** (JDK 21):

  | payloadSize | throughput |
  |---|---|
  | 32 | 94.3 ± 2.6 ops/s |
  | 256 | 94.0 ± 3.1 ops/s |
  | 2048 | 91.1 ± 12.3 ops/s |

  Quick config (1 fork / 1 warmup / 3 iters). Flat across payload sizes — the round-trip is dispatch/`runBlocking`-bound, not payload-bound — with no errors (`-foe true`). It measures the same operation as before, so no benchmark-semantics regression.

Non-published module; no API/ABI impact.


monkopedia-reviewer (approved):
Reviewed and approved.

**Fix is correct:** `KsrpcNativeHost.connect(scope, TestNativeHost::initialize, env)` is the right API — confirmed against `KsrpcNativeHost.kt` (3-arg form; `environment` has a default). The `@CName("Java_com_monkopedia_ksrpc_TestNativeHost_initialize")` in `ksrpc-test/src/nativeMain/kotlin/Test.kt` resolves the new `object TestNativeHost` binding correctly (package `com.monkopedia.ksrpc`, name mangling matches).

**NativeHost deletion verified:** No remaining references in `ksrpc-bench/src`. The `NativeHost` that still exists in `ksrpc-test` is a separate class with distinct native bindings for low-level JNI tests — unaffected.

**Benchmark semantics unchanged:** Only the untimed `@Setup` path changed; the `@Benchmark fun jniRpcRoundTrip()` is untouched.

**CI guard is the right minimal prevention:** `:ksrpc-bench:compileKotlinJvm` in jvm-tests catches future drift without running JMH.

All 5 CI checks green. Code is idiomatic and consistent with surrounding patterns.

Issue #246, opened 2026-07-15
CI never runs kotlinWasmNpmInstall — wasm npm/engine breakage ships undetected


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)
**Category:** test-gap / CI

**What happened:** 1.1.4 bumped npm `nanoid` to 6.0.0, whose `engines.node = "^22 || ^24 || >=26"` excludes Node 25 — the default managed Node that Kotlin 2.4.10 provisions for the Wasm/JS toolchain. This broke `:kotlinWasmNpmInstall` for downstream consumers doing a full Wasm build, but **passed all of ksrpc's own CI green** and only surfaced when a consumer (lsp-kotlin) hit it. Fixed in 1.1.5 (revert to nanoid 5.1.16).

**The gap:** ksrpc CI never runs `kotlinWasmNpmInstall` (or `kotlinNpmInstall`). The lint job's `apiCheck` builds klibs, which don't npm-link; `compileKotlinWasmJs` compiles the external binding without resolving npm. So any bad npm dependency or `engines` incompatibility passes CI and ships to consumers.

**Proposed:** add a CI step that actually resolves the js/wasm npm graph, e.g. `./gradlew kotlinNpmInstall kotlinWasmNpmInstall` (or a light `jsNodeTest`/`wasmJsNodeTest` if a node test runner is wanted). This runs on ubuntu (KGP provisions Node), so it fits the existing default CI — the cost is the Node download + a yarn resolve.

**Why:** catches npm-dep/engine breakage at PR time instead of in a consumer's build. This is the js/wasm analog of the `ci-apple` label job (#241) — those targets aren't exercised by default CI.

**Effort:** small (one CI step). **Decision needed:** whether to always-run it in the default jobs vs. gate it (label/`workflow_dispatch`) to save the Node-provision cost on unrelated PRs.

_Filed after the 1.1.4→1.1.5 nanoid/Node-25 incident._

monkopedia-coder:
A concrete instance for this, from #247 / #248.

`asKsFlow()` was returning its collection token before the source was actually subscribed, so a hot source lost everything emitted in that window. Measured on the JVM at 1877 drops per 2000 rounds. On JS and wasmJs it is worse: those targets are single-threaded, so the scheduled collection cannot possibly run before the caller resumes — there is no second thread to win the race, which is exactly what the 6% of surviving JVM rounds depended on. The defect is total there, not partial.

[...]

PR #248, opened 2026-08-01
Subscribe to the source before startCollection returns (#247)

### LINKED ISSUE #247: asKsFlow: startCollection returns before the source is subscribed, dropping hot-source emissions
`asKsFlow()` starts its collection with `scope.launch { source.collect { ... } }` and returns the `KsCollectionToken` immediately. `launch` (default `CoroutineStart.DEFAULT`) only *schedules* the body, so `startCollection` can return to the caller before the coroutine has subscribed to `source`.

For a **cold** source this is harmless — nothing emits until the collection starts, so nothing can be missed. For a **hot** source (a `MutableSharedFlow`/`SharedFlow` with `replay = 0`, or any flow the user shares) every emission in that window is dropped, and the client sees a stalled collection rather than an error.

This matters because it silently breaks the one handshake callers have. `startCollection` returning a token is the only "I am subscribed now" signal ksrpc offers, and today it does not mean that.

### Measurement

In-process stress against `asKsFlow(MutableSharedFlow(replay = 0))`, 2000 rounds, `Dispatchers.Default`: `startCollection(collector)` then `emit(1)` immediately, 200 ms per-round timeout.

| variant | drops |
| --- | --- |
| current (`launch { }`), emit right after the token returns | 1877 / 2000 |
| control: await `subscriptionCount > 0` before emitting | 0 / 2000 |
| `launch(start = CoroutineStart.UNDISPATCHED)`, emit right after the token returns | 0 / 2000 |

The control isolates the cause: the drop is entirely the gap between the token returning and the launched body subscribing. In-process the whole window is that dispatch gap, which is why the rate is so high; over a real connection the trigger takes a wire round trip, so the rate falls but the window does not close, and it widens under dispatcher load.

### Fix

`scope.launch(start = CoroutineStart.UNDISPATCHED)`. The body then runs on the calling coroutine up to its first real suspension, which for `SharedFlow.collect` is after the subscriber slot is registered — so `startCollection` cannot return before the subscription exists. `onSubscription` is not usable here: it is defined on `SharedFlow`, and `asKsFlow` accepts any `Flow`.

The trade-off is that a cold source's synchronous prologue (the part of a `flow { }` builder before its first suspension) now runs on the caller's dispatcher instead of the launched one. In practice the first `collector.onItem` is a suspending sub-service call, so a source that emits at all suspends almost immediately.

Private implementation detail — no ABI change, nothing in the api dumps moves. The 34 existing jvm flow tests (`FlowAutoDetectTest`, `KsFlowServiceTest`, `FlowIntrospectionTest`, `NestedGenericIntrospectionTest`, `KsServiceHierarchyTest`) pass with the change.

Should also come with a regression test pinning the hot-source case, and a KDoc line on `asKsFlow` stating that `startCollection` returning means the source is subscribed.

monkopedia-coder:
Fixes #247.

`asKsFlow()` started its server-side collection with a plain `scope.launch`, which only *schedules* the body, and returned the `KsCollectionToken` right away. So `startCollection` could return before the coroutine had subscribed to the source.

A cold source is unaffected — nothing emits until the collection starts. A hot source (a `MutableSharedFlow` with `replay = 0`, or any flow the user shares) drops everything emitted in that window, and the client sees a stalled collection rather than an error. That matters because `startCollection` returning a token is the only "I am subscribed now" signal ksrpc gives a caller, and it did not mean that.

### Fix

`scope.launch(start = CoroutineStart.UNDISPATCHED)`. The body runs on the calling coroutine up to its first real suspension, which for `SharedFlow.collect` is after the subscriber slot is registered — so the token cannot be handed back before the subscription exists.

`onSubscription` is not usable here: it is defined on `SharedFlow`, and `asKsFlow` accepts any `Flow`.

The trade-off is that a cold source's synchronous prologue (the part of a `flow { }` builder before its first suspension) now runs on the caller's dispatcher instead of the launched one. In practice the first `collector.onItem` is a suspending sub-service call, so a source that emits at all suspends almost immediately.

Private implementation detail — no ABI change, `apiCheck` is byte-identical.

### Measurement

2000 rounds in-process on `Dispatchers.Default`, `startCollection(collector)` then `emit` immediately, 200 ms per round:

| variant | drops |
| --- | --- |
| before (`launch { }`) | 1877 / 2000 |
| control: await `subscriptionCount > 0` before emitting | 0 / 2000 |
| after (`launch(start = UNDISPATCHED)`) | 0 / 2000 |

The control isolates the cause to the dispatch gap. In-process the whole window *is* that gap, which is why the rate is so high; over a real connection the trigger costs a wire round trip, so the rate falls but the window does not close, and it widens under dispatcher load.

### Test

`KsFlowServiceTest.testStartCollectionSubscribesBeforeReturning` asserts `subscriptionCount.value == 1` the instant `startCollection` returns, then emits and requires the item to arrive. It asserts the invariant itself rather than sampling for the symptom, so there are no sleeps or timing thresholds.

It repeats 100 rounds, because a scheduled collection can still win the race by luck — it subscribed in time in about 6% of the measured rounds (the 123 of 2000 that did not drop), so a single round would let the regression through about that often. 100 rounds run in ~1.2s. Verified to fail on the unfixed code (`round N: startCollection returned before subscribing to the source expected:<1> but was:<0>`) and pass with the change.

### Verification

`ktlintCheck` (CI's `-x ktlintKotlinScriptCheck` form), `licenseCheckForKotlin`, `apiCheck`, full `jvmTest`, and `:ksrpc-test:linuxX64Test` all green locally on JDK 21.

### DIFF
--- a/ksrpc-flow/src/commonMain/kotlin/com/monkopedia/ksrpc/flow/AsKsFlow.kt
+++ b/ksrpc-flow/src/commonMain/kotlin/com/monkopedia/ksrpc/flow/AsKsFlow.kt
@@ -20,6 +20,7 @@ import com.monkopedia.ksrpc.asString
 import kotlin.coroutines.coroutineContext
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.FlowCollector
@@ -34,6 +35,10 @@ import kotlinx.coroutines.launch
  * [KsFlowCollector]. The returned [KsCollectionToken] can cancel that
  * collection.
  *
+ * The collection is subscribed to this flow before `startCollection` returns,
+ * so a hot source (a `SharedFlow` with no replay, for instance) can emit as
+ * soon as the caller holds the token without those emissions being missed.
+ *
  * The [KsFlowService] does NOT auto-close on collection completion — it
  * stays alive until [KsFlowService.close] is called.
  */
@@ -44,7 +49,10 @@ private class KsFlowServiceImpl<T>(private val source: Flow<T>) : KsFlowService<
 
     override suspend fun startCollection(collector: KsFlowCollector<T>): KsCollectionToken {
         val scope = CoroutineScope(coroutineContext + parentJob)
-        val job = scope.launch {
+        // UNDISPATCHED so the collection subscribes to [source] before this call
+        // returns; a plain launch only schedules the body, and a hot source drops
+        // everything emitted between the token returning and the body running.
+        val job = scope.launch(start = CoroutineStart.UNDISPATCHED) {
             try {
                 source.collect { item ->
                     collector.onItem(item)
--- a/ksrpc-test/src/commonTest/kotlin/KsFlowServiceTest.kt
+++ b/ksrpc-test/src/commonTest/kotlin/KsFlowServiceTest.kt
@@ -33,6 +33,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.FlowCollector
+import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.toList
 import kotlinx.coroutines.launch
@@ -236,6 +237,47 @@ class KsFlowServiceTest {
             )
         }
     )
+
+    /**
+     * A hot source must be subscribed by the time `startCollection` returns.
+     * If the collection is only scheduled, everything the source emits between
+     * the token returning and the collection starting is dropped and the
+     * collector waits out its timeout instead (#247).
+     *
+     * Repeated because a scheduled collection can still win the race by luck —
+     * it subscribed in time in roughly 6% of rounds when this was measured, so
+     * a single round would let the regression through about that often.
+     */
+    @Test
+    fun testStartCollectionSubscribesBeforeReturning() = runBlockingUnit {
+        repeat(100) { round ->
+            val source = MutableSharedFlow<String>(replay = 0)
+            val flowService = source.asKsFlow()
+            val received = CompletableDeferred<String>()
+            val collector = object : KsFlowCollector<String> {
+                override suspend fun onItem(item: String) {
+                    received.complete(item)
+                }
+
+                override suspend fun onError(error: RpcFailure) = Unit
+
+                override suspend fun onComplete() = Unit
+
+                override suspend fun close() = Unit
+            }
+            flowService.startCollection(collector)
+            assertEquals(
+                1,
+                source.subscriptionCount.value,
+                "round $round: startCollection returned before subscribing to the source"
+            )
+            source.emit("emitted-right-after-token")
+            withTimeout(5000) {
+                assertEquals("emitted-right-after-token", received.await())
+            }
+            flowService.close()
+        }
+    }
 
     private fun executePipe(
         serviceJob: suspend (Connection<String>) -> Unit,


monkopedia-coder:
One open question for whoever reviews this: does the JS/wasm side matter for a dispatch-timing change? Answering it here so it doesn't have to be rediscovered.

It matters, and it argues for the fix rather than against it. `CoroutineStart.UNDISPATCHED` is a coroutines-core semantic with the same meaning on every target — it is not a JVM threading trick. On JS and wasmJs the situation before this change is strictly worse than on the JVM: those targets are single-threaded, so a `CoroutineStart.DEFAULT` launch cannot run before the caller resumes. There is no other thread to win the race, so the collection is guaranteed not to have subscribed when `startCollection` returns, and a hot source would drop **every** emission rather than the 94% measured on `Dispatchers.Default`. The 6% of JVM rounds that got away with it are exactly the ones where another thread happened to run the body in time — a thing that cannot happen there.

The regression test lives in `commonTest`, so it covers those targets whenever their test tasks are run. CI is ubuntu-only and does not run the JS/wasm suites (see #246), so this is not verified by the checks on this PR; `:ksrpc-test:linuxX64Test` green does cover a native target.

monkopedia-reviewer (approved):
Reviewed head `344e0c66325ec41866b1135efc48db3b46cf0e75`.

**Mechanism.** `scope.launch(start = CoroutineStart.UNDISPATCHED)` runs the coroutine body synchronously on the caller's thread up to its first real suspension. For `SharedFlow.collect`, the subscriber slot is registered before any suspension, so the token cannot be returned before the subscription exists — this checks out against `kotlinx.coroutines`' actual `SharedFlowImpl` contract, not just the PR's description of it.

**Regression test.** `testStartCollectionSubscribesBeforeReturning` asserts `subscriptionCount.value == 1` the instant `startCollection` returns — an invariant assertion, not a timing sample, so no sleeps/thresholds to go flaky. On the pre-fix code this assertion is genuinely racy rather than deterministically false (a scheduled `launch{}` can still win the race if another thread services it before the assertion runs), which is exactly why 100 rounds instead of 1 — the PR's own measurement (123/2000 escapes, ~6%) justifies the repeat count rather than it being a round number picked for comfort. This is a test that can fail against the bug it guards, not decoration.

**The UNDISPATCHED trade-off.** The PR states the cost as "a cold source's synchronous prologue runs on the caller's dispatcher instead of the launched one," reasoning that in practice `collector.onItem` is a suspending sub-service call so a source that emits at all suspends almost immediately. I checked this against the actual call chain rather than taking it on faith: `KsFlowCollector.onItem` is a real `@KsMethod` RPC call (not `@KsNotification`), and on the server side the `collector` passed into `KsFlowServiceImpl.startCollection` is always the compiler-generated wire stub — never user code — so it does suspend on essentially every call in the shipped path. I also checked whether running the prologue inline could disturb wire ordering (e.g. an `onItem` call going out before the `/collect` response that returns the token) — `MultiChannel` multiplexes pending calls by id with no FIFO assumption, so out-of-order bidirectional traffic is already a normal, supported shape in this transport, not something this change introduces. The one place the "prologue only" framing doesn't fully hold is the advanced/raw `KsFlowService` API used directly in-process with a hand-written non-suspending collector (as the test itself does) — there, a fully eager cold source could run to completion inline before the token is returned. That's a real but narrow corner (test/local-composition usage, not the wire path), not worth blocking on.

**KDoc.** The added paragraph ("subscribed to this flow before startCollection returns... a hot source... can emit as soon as the caller holds the token without those emissions being missed") matches what the code guarantees for the documented/common case.

**JS/wasm.** The PR's own follow-up comment correctly identifies that JS/wasm are single-threaded, so the pre-fix bug there is a 100% drop rather than the 94% measured on `Dispatchers.Default` — and correctly discloses that CI doesn't run those suites (tracked separately in #246), so this fix is unverified-not-verified-good on that platform. That's the right way to state it.

**Mechanics.** Private impl detail (`KsFlowServiceImpl` is `private`), no ABI movement — confirmed independently via the `lint` job's `apiCheck` pass, not just the claim. All 5 required checks (compiler-tests, jni-tests, jvm-tests, lint, native-tests) pass on this head; `apple` is label-gated and not required. No `GlobalScope`, no scope creep, comment/KDoc style matches the file's existing conventions.

Approving and merging.

Issue #255, opened 2026-08-07
First-ever jsTest run: 371 of 383 pass, 12 fail across four classes


monkopedia-coder:
The `jsTest` suite has never been executed by any workflow (#251). Ran it for the first time on 2026-08-07 against `ChromeHeadless 150` via the configured karma launcher:

```
:ksrpc-test:jsBrowserTest — 383 tests, 12 failures, 0 errors, 0 skipped   (45s)
```

**371 of 383 pass**, so the suite is broadly healthy and the js target is not fundamentally broken. Twelve fail, in four classes, with two distinct signatures:

| class | failing | signature |
| --- | --- | --- |
| `NestedGenericIntrospectionTest` | 6 | `ClosedWriteChannelException: Throwable` |
| `PacketCancellationTest` | 3 | `ClosedWriteChannelException: Throwable` |
| `MultiChannelRaceTest` | 1 | `ClosedWriteChannelException: Throwable` |
| `JsonRpcTierCheckTest` | 2 | `Error: Timeout of 2000ms exceeded` |

Ten of the twelve share one exception, which suggests a single cause in how the js `createPipe()` / `ByteChannel` pairing behaves under the browser event loop rather than ten independent defects — but **that is a hypothesis from the grouping, not something I have traced.** Nobody should treat the "one root cause" reading as established until someone follows one of them into the transport.

Two things this does not tell us:

- **Whether these are defects in shipped js code or artifacts of the test harness on this platform.** `createPipe()` in `jsTest/TestUtilsJs.kt` returns the same `ByteChannel` for both ends, and a browser event loop schedules differently from a JVM dispatcher; a close that is benign on the JVM may land differently here. Either answer is worth knowing and they need different fixes.
- **Whether they are new.** These tests have never run on js, so there is no "it used to pass" to compare against. The failures could predate any recent change by years.

The `JsonRpcTierCheckTest` timeouts are a separate signature and may be a genuinely slow path rather than a hang — the karma default is 2000ms, which is tight for a suite that stands up a transport.

**Reproduce:**

```bash
CHROME_BIN=/usr/bin/chromium ./gradlew :ksrpc-test:jsTest
```

Chromium must be on the box and `CHROME_BIN` set — the karma config uses `useChromeHeadless()` and does not locate a browser itself. That is also a prerequisite for any CI job that runs this suite.

Blocks the js half of #251: wiring this into CI means either fixing these twelve first or landing the job with them known-failing, and the second is not a real option.

[...]

PR #257, opened 2026-08-07
Fix the wasmJsTest compile break so the suite can run at all (#256)

### LINKED ISSUE #256: wasmJsTest source set does not compile — promise block must return JsAny? on wasmJs
The `wasmJsTest` source set does not compile. Found while running the suites for the first time (#251):

```
e: ksrpc-test/src/wasmJsTest/kotlin/TestUtilsJs.kt:100:24 Cannot infer type for type parameter 'T'. Specify it explicitly.
e: ksrpc-test/src/wasmJsTest/kotlin/TestUtilsJs.kt:101:9 Return type mismatch: expected 'JsAny?', actual 'Unit'.
```

The code is the wasm `actual` for the shared test runner:

```kotlin
internal actual fun runBlockingUnit(
    function: suspend CoroutineScope.() -> Unit
): RunBlockingReturn {
    @Suppress("CAST_NEVER_SUCCEEDS")
    return GlobalScope.promise {
        function()
    } as RunBlockingReturn
}
```

On wasmJs, `promise`'s block must produce a `JsAny?`; the test body returns `Unit`, so `T` cannot be inferred. The js `actual` next door is unaffected because it goes through `asDynamic()`.

**This is the ksrpc-bench pattern from #242 again: code that nothing compiles rots against the toolchain.** The wasm test sources are not built by `apiCheck` (which compiles `wasmJsMain`, not `wasmJsTest`), and no workflow runs the suite, so a Kotlin upgrade tightened wasm interop typing and nothing noticed. The nearest candidate is the 2.4.0 → 2.4.10 bump in 1.1.4, but I have not bisected it and the break may be older.

The fix is small — give `promise` an explicit type argument and a `null` last expression:

```kotlin
return GlobalScope.promise<JsAny?> {
    function()
    null
} as RunBlockingReturn
```

**That edit compiles in principle and is NOT yet verified** — the build slot went to another agent before I could run it, and I am not opening a PR for an unverified fix. What is verified is the break itself, quoted above.

Note the sequencing: until this compiles, **nobody knows anything about the wasm suite's actual pass rate.** The js half of #251 came back 371/383 (#255); the wasm half is still entirely unmeasured, and the compile break is why.

**Reproduce:** `CHROME_BIN=/usr/bin/chromium ./gradlew :ksrpc-test:wasmJsTest`

monkopedia-coder:
Fixes #256.

The wasmJs test source set did not compile:

```
e: ksrpc-test/src/wasmJsTest/kotlin/TestUtilsJs.kt:100:24 Cannot infer type for type parameter 'T'.
e: ksrpc-test/src/wasmJsTest/kotlin/TestUtilsJs.kt:101:9 Return type mismatch: expected 'JsAny?', actual 'Unit'.
```

`promise`'s block must yield a `JsAny?` on wasmJs; the test body returns `Unit`, so `T` could not be inferred. The js actual next door is unaffected because it goes through `asDynamic()`.

Nothing in this repo compiles the wasm test sources — `apiCheck` builds `wasmJsMain`, not `wasmJsTest`, and no workflow runs the suite (#251) — so a toolchain tightening broke them and nothing noticed. The nearest candidate is the Kotlin 2.4.0 → 2.4.10 bump in 1.1.4, but I have not bisected it and the break may be older. Same shape as the ksrpc-bench rot in #242: code that nothing compiles rots against the toolchain.

### What this does and does not do

**Does:** make `:ksrpc-test:wasmJsTest` compile and execute. Verified — the suite now runs 386 tests where previously it could not report anything at all.

**Does not:** make it pass. **46 of those 386 fail**, which is #251's problem, not this PR's. Breakdown for whoever picks that up:

| count | signature |
| --- | --- |
| 33 | `IllegalStateException: Can't find rpc companion for class …` |
| 10 | `ClosedWriteChannelException` |
| 2 | karma `Timeout of 2000ms exceeded` |
| 1 | *(was mine — a #254 regression, already reverted there)* |

The 10 + 2 match the js run's signatures exactly (#255), so those two shapes are not wasm-specific. The 33 companion-lookup failures are wasm-only and are the interesting ones.

I am deliberately not bundling any of that here. This PR is the smallest change that turns "we cannot know" into "we know, and it is 46."

**Verification:** `CHROME_BIN=/usr/bin/chromium ./gradlew :ksrpc-test:wasmJsTest` compiles and runs (JDK 21, headless Chromium 150). The suite exits non-zero on those 46 failures, so CI would be red if this were wired in today — which is exactly why #251 stays open rather than being closed by this.

### DIFF
--- a/ksrpc-test/src/wasmJsTest/kotlin/TestUtilsJs.kt
+++ b/ksrpc-test/src/wasmJsTest/kotlin/TestUtilsJs.kt
@@ -96,8 +96,11 @@ actual typealias RunBlockingReturn = Any
 internal actual fun runBlockingUnit(
     function: suspend CoroutineScope.() -> Unit
 ): RunBlockingReturn {
+    // promise's block must produce a JsAny? on wasmJs, so the Unit-returning
+    // test body cannot be the last expression.
     @Suppress("CAST_NEVER_SUCCEEDS")
-    return GlobalScope.promise {
+    return GlobalScope.promise<JsAny?> {
         function()
+        null
     } as RunBlockingReturn
 }


monkopedia-reviewer (approved):
Independent verification, not just a diff read.

**Pre-fix claim** — reproduced exactly. Swapped in `main`'s version of the file and ran `./gradlew :ksrpc-test:compileTestKotlinWasmJs`:
```
e: TestUtilsJs.kt:100:24 Cannot infer type for type parameter 'T'. Specify it explicitly.
e: TestUtilsJs.kt:101:9 Return type mismatch: expected 'JsAny?', actual 'Unit'.
```
Same file, same lines, same messages as the PR body. This establishes the "did not compile at all" claim as fact, not assertion.

**Post-fix claim** — compiles and runs; count is close but not exact. `CHROME_BIN=/usr/bin/chromium ./gradlew :ksrpc-test:wasmJsTest` on this head: **385 tests, 47 failed** (33 `IllegalStateException` companion-lookup, 12 `ClosedWriteChannelException`, 2 karma timeouts) vs. the PR's claimed **386/46** (33/10/2/1). The delta is 2 extra `ClosedWriteChannelException` failures and one fewer completed test — plausible run-to-run flakiness given the failing tests are named `MultiChannelRaceTest`/`PacketCancellationTest`/race-teardown-shaped, not a fabricated number. Worth a one-line ack in the PR body that the exact count can drift a test or two between runs, but it doesn't change the substance: the suite goes from "cannot report anything" to "runs and reports N in the same ballpark," which is the actual claim being made.

**CI-green caveat** — confirmed independently by reading `.github/workflows/ci.yaml`: `lint`/`jvm-tests`/`jni-tests`/`native-tests`/`compiler-tests` are the only jobs, none invoke `wasmJsTest` or even `jsTest`. `apiCheck` only runs `compileKotlinWasmJs` (main, not test). So this PR's green checks are silent on whether the change works — that's the point of #251, and the PR body is honest about it rather than leaning on the green to imply more than it shows.

**Scoping legitimacy** — #251, #255, #258 all exist, are open, and #258 in particular is not a stub: it already has the 33/10/2 breakdown by test class and states the companion-lookup hypothesis without overclaiming a diagnosis. The scoping-out of the 46 failures is real, not a way to sneak a red suite past review — nothing here wires this into CI, so nothing goes red as a result of merging.

**Fix quality** — the explicit `<JsAny?>` type argument plus trailing `null` is the natural fix for `kotlinx.coroutines.promise`'s `T : JsAny?` bound on wasmJs; comparing to `jsTest`'s `asDynamic()`-based actual, the two backends need different treatment here (dynamic sidesteps the bound entirely on legacy JS; wasmJs enforces it statically), so this isn't a case of the wasm actual reinventing a pattern the js actual already solved better. `@Suppress("CAST_NEVER_SUCCEEDS")` and the `as RunBlockingReturn` cast are unchanged in effect — the cast is from an external Promise-interop type to `Any`, which the compiler still can't statically justify, and the underlying runtime object is unaffected by giving `T` an explicit value. No hidden risk introduced. Diff is minimal, test-only (`internal actual fun` in a test source set — no public API, no production code), comment is accurate and non-redundant.

**Classification**: `test_only`, project default tier 1 (no override for this category on ksrpc). No `Monkopedia` reviewer request on the PR, nothing performance- or API-adjacent, no new `GlobalScope` use (pre-existing, and it's test code — the ksrpc `GlobalScope` flag targets non-test code).

Verdict: **approve**. Correct, minimal, honestly scoped, and independently reproduced on both ends (break and fix). The only nit is the off-by-a-couple test count in the table, not blocking.

Per instruction from whoever spawned this review, I am not merging — leaving that to the coordinator/owner.

Issue #258, opened 2026-08-07
33 wasmJs tests fail with "Can't find rpc companion for class" — companion lookup may not resolve on wasm


monkopedia-coder:
The wasmJs suite runs for the first time once #256's compile break is fixed (PR #257). 386 tests, 46 failures — and **33 of them are one signature that does not appear on any other platform**:

```
kotlin.IllegalStateException: Can't find rpc companion for class com.monkopedia.…
```

Distribution across classes:

| class | failures |
| --- | --- |
| `MethodMetadataArgShapesTest` | 13 |
| `MethodMetadataPropagationTest` | 12 |
| `NestedGenericIntrospectionTest` | 6 |
| `GenericServiceSubtypeTest` | 4 |
| `TimedGenericMetadataTest` | 4 |

(The counts include a handful of `ClosedWriteChannelException` failures in `NestedGenericIntrospectionTest`, which are the js-shared signature tracked in #255 — the companion-lookup subtotal is 33.)

**Why this one matters more than the raw count.** The other two wasm signatures (`ClosedWriteChannelException` ×10, karma timeouts ×2) appear with identical shapes in the js run, so they are not wasm-specific. This one is. The affected classes are concentrated in metadata, introspection and generic-service resolution, which is where the compiler plugin's generated `RpcObject`/companion registration is consulted — so the hypothesis worth testing first is that **companion lookup does not resolve on wasmJs the way it does on js and jvm**, rather than 33 unrelated defects.

**That is a hypothesis, not a diagnosis. Nobody has traced one of these to its cause yet**, and "same message, same area" is exactly the reasoning that produces confident wrong groupings. The first useful piece of work is following a single failure — `MethodMetadataArgShapesTest` is the densest and probably the cheapest — into the lookup and finding out whether the companion is missing, registered under a different key, or present and unreachable.

**What this means for shipped artifacts, stated carefully.** ksrpc publishes `-wasm-js` coordinates for every module. If the lookup genuinely fails at runtime on wasmJs rather than only under the test harness, then service introspection and metadata do not work for wasm consumers today. **I have not confirmed which**, and the distinction is the whole question: a harness-only artifact is a test problem, a runtime failure is a shipped defect.

**Reproduce:**

```bash
# with PR #257 applied, or the equivalent TestUtilsJs fix
CHROME_BIN=/usr/bin/chromium ./gradlew :ksrpc-test:wasmJsTest
```

[...]

PR #259, opened 2026-08-07
SPIKE / DO NOT MERGE: measure the js and wasm suites on a CI runner (#251)

monkopedia-coder:
**⚠️ SPIKE — DO NOT MERGE.** This exists to produce a measurement, not to land a change. For #251.

Everything known about ksrpc's js and wasmJs suites comes from **one machine, mine**:

| | js | wasmJs |
| --- | --- | --- |
| run 1 (mine) | 383 tests, 12 failures | 386 tests, 46 failures |
| run 2 (a reviewer's, same code) | — | 385 tests, **47** failures |

The count that moved is `ClosedWriteChannelException`, in race- and teardown-shaped tests. So part of what we "know" about these suites is one sample of a distribution, measured on a loaded developer box running fifteen agents.

### What this job answers that a local run cannot

1. **Does karma find a browser on `ubuntu-latest` without `CHROME_BIN` set by hand?** Locally it does not — the config uses `useChromeHeadless()` and does not locate a browser itself, so I had to point `CHROME_BIN` at `/usr/bin/chromium`. Whether a GitHub image needs the same is a prerequisite for any real job and nobody has checked it. The job prints browser availability *before* running anything, so the answer is visible even if everything after it fails.
2. **Do the failures reproduce on a clean machine, or are some of them mine?** A CI runner is a different scheduler, a different browser build, and no competing load.
3. **What does it cost in wall-clock?** That is the actual input to the always-run vs label-gate decision in #251, and I have been guessing at it from a local 45s.

It also prints **executed test counts per source set**, because a green with zero executions and a green with 383 executions are otherwise indistinguishable — five separate mechanisms for that were found across the fleet this evening.

### Why this must not merge as-is

It carries `continue-on-error: true`. **This repo has no other instance of that flag** — I checked all five workflows during an audit today, precisely because a verification step that cannot fail is a known trap. Here it is deliberate and load-bearing for a measurement (the suites are known-failing; the measurement must not block the rest of CI), and it is exactly the wrong thing to leave in a real job.

So: read the run, take the numbers, close this. The real job belongs in whatever addresses #251, without this flag, and only after #255 / #258 are resolved.

### The case against doing this at all

If the CI numbers come back matching mine, this branch bought a confirmation and cost a workflow file. That is the likely outcome and it is still worth the runner minutes, because the browser-availability question is a genuine unknown that blocks writing the real job — and it is cheaper to learn it here than inside a PR that is trying to fix something.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -174,9 +174,26 @@ jobs:
       - name: Summarise executed counts
         if: always()
         run: |
-          for d in ksrpc-test/build/test-results/*/; do
-            n=$(grep -ho 'tests="[0-9]*"' "$d"*.xml 2>/dev/null | grep -o '[0-9]*' | paste -sd+ | bc || echo 0)
-            f=$(grep -ho 'failures="[0-9]*"' "$d"*.xml 2>/dev/null | grep -o '[0-9]*' | paste -sd+ | bc || echo 0)
+          # An absent results directory and a directory of empty results must not
+          # look the same, and neither may print an empty number: this step exists
+          # to distinguish "ran nothing" from "ran and passed", so it states 0
+          # explicitly and says NO-RESULTS-DIR when there is nothing to read.
+          # Uses sed rather than grep because grep matching nothing exits 1, which
+          # under a stricter shell would redden the job with a misleading message.
+          shopt -s nullglob
+          dirs=(ksrpc-test/build/test-results/*/)
+          if [ ${#dirs[@]} -eq 0 ]; then
+            echo "NO-RESULTS-DIR — no test task produced results; executed=0"
+            exit 0
+          fi
+          for d in "${dirs[@]}"; do
+            xmls=("$d"*.xml)
+            if [ ${#xmls[@]} -eq 0 ]; then
+              echo "$d executed=0 failures=0 (no XML in directory)"
+              continue
+            fi
+            n=$(sed -n 's/.*tests="\([0-9]*\)".*/\1/p' "${xmls[@]}" | awk '{s+=$1} END {print s+0}')
+            f=$(sed -n 's/.*failures="\([0-9]*\)".*/\1/p' "${xmls[@]}" | awk '{s+=$1} END {print s+0}')
             echo "$d executed=$n failures=$f"
           done
       - name: Upload results
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -140,3 +140,48 @@ jobs:
           :ksrpc-binary-okio:jvmJar
       - name: Run compiler plugin tests
         run: cd compiler && ./gradlew test
+
+  # SPIKE / DO NOT MERGE — measurement job for #251.
+  # The js and wasmJs suites have never run in CI. This job exists to answer
+  # three questions on a clean GitHub runner rather than one developer's box:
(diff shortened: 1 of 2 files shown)


monkopedia-coder:
**First results, and the js half answered the question that actually blocked writing a real job.**

```
js:    executed=379  failures=12          <- ran on the runner
wasm:  compileTestKotlinWasmJs FAILED     <- blocked on #257, not a measurement
```

**1. karma found a browser with no `CHROME_BIN` and no setup step.** That was the genuine unknown — locally the suite cannot start without pointing `CHROME_BIN` at a chromium binary, and whether a GitHub image needs the same had never been checked. It does not. **So the real job needs no browser provisioning**, which is the single most useful thing this spike was for.

**2. The js failure count is identical to local; the executed count is not.**

| | executed | failures |
| --- | --- | --- |
| local (my box) | 383 | 12 |
| CI runner | 379 | 12 |

Twelve failures both times, four fewer tests executed on CI. That is worth a second look rather than a shrug — `platformSupportedTestTypes()` in `jsTest/TestUtilsJs.kt` adds `SERVICE_WORKER` only `if (hasWindow())`, so a headless-Chrome-without-window difference between the two environments is the obvious candidate for tests silently not being generated. **Same failures, fewer executions, and only the executed count reveals it** — which is exactly the property the count summary in this job exists to surface.

**3. The wasm half is blocked, not broken.** This branch was cut from `main`, which does not carry #257's compile fix, so `compileTestKotlinWasmJs` failed exactly as it does on `main`. **That is a dependency the spike surfaced rather than a result it produced** — and it independently confirms #257's premise from a clean machine: the wasm test sources genuinely do not compile on `main`.

I have cherry-picked #257's commit onto this branch so the wasm half can measure. **That makes this branch carry someone else's pending fix, which is fine for a spike and is another reason it must not merge.**

Still unanswered until the next run: wall-clock for the wasm half, and whether the wasm failure count on a clean machine matches either of the two local samples (386/46 and 385/47).

monkopedia-coder:
Closing this. It was a measurement, it produced its measurements, and leaving it open now costs more than it returns — spike branches are triggering HTTP 429s from Maven Central across the fleet, and a draft that re-runs a two-suite browser job on every push is exactly the wrong thing to keep alive.

**What it answered, which is why it existed:**

1. **karma finds a browser on `ubuntu-latest` with no `CHROME_BIN` and no setup step.** This was the genuine unknown and it blocked writing a real job — locally the suite cannot start without pointing `CHROME_BIN` at a chromium binary. The eventual job in #251 needs no browser provisioning.
2. **js on a clean runner: 379 executed, 12 failures** — same failure count as my box, **four fewer tests executed**. That discrepancy became #261 (tests that early-return report as PASSED, so the executed count overstates coverage) and is a better finding than the number itself.
3. **The wasm half was blocked, not broken** — this branch predated #257, so `compileTestKotlinWasmJs` failed exactly as it did on `main`, independently confirming #257's premise from a machine that was not mine.

**What it did not answer:** wasm counts and wall-clock on a clean runner. That is now cheap to redo whenever it matters, because **#257 has merged**, so `main` compiles the wasm test sources — which was the whole reason this branch had to carry a cherry-pick.

**Never merge this branch.** It holds `continue-on-error: true`, the only instance in the repo, plus a duplicate of #257's commit. Both were deliberate for a measurement and wrong for anything else.

Everything above is recorded in #251, #255, #258 and #261, so closing this loses nothing.

PR #260, opened 2026-08-07
Send the flow error's message, not the server's stack trace (#250)

### LINKED ISSUE #250: Flow errors send the server's full stack trace to the client, contrary to the documented convention
The main call path deliberately keeps server stack traces off the wire, and says so:

`ksrpc-core/src/commonMain/kotlin/internal/HostSerializedServiceImpl.kt:113-118`

```kotlin
// Concise message only — the full stack is logged server-side via env.logger
// above, not propagated to the client. Matches RpcMethod.encodeError's
// convention so wire frames are consistent regardless of which catch
// produced them.
CallData.Error(code, t.message ?: t.toString())
```

`RpcMethod.encodeError` (`RpcMethod.kt:524`) does the same — `t.message ?: t.toString()`.

The flow path does not. `ksrpc-flow/src/commonMain/kotlin/com/monkopedia/ksrpc/flow/AsKsFlow.kt:65`:

```kotlin
collector.onError(RpcFailure(e.asString))
```

`asString` is the full stack trace, not the message:

- `ksrpc-core/src/jvmMain/kotlin/RpcObject.kt:77-80` — `StringWriter().also { printStackTrace(PrintWriter(it)) }.toString()`
- `ksrpc-core/src/nativeMain/kotlin/RpcObject.kt:72-73` — `message + "\n" + getStackTrace().joinToString("\n")`
- `ksrpc-core/src/jsAndWasmJsMain/kotlin/RpcObject.kt:65-66` — `toString()` (message only; JS/wasm hosts are unaffected)

`collector` is the client's `KsFlowCollector` sub-service, so the string crosses the connection. On the client, `KsFlowService.collect` turns it into `RpcException(stack)` via `RpcFailure.toException()`.

So on a JVM or Native host, any exception thrown by a flow that a client is collecting sends that client the server's full stack trace: internal class and package names, source file names and line numbers, and the frames of whatever libraries are on the server's stack. The main call path sends only `t.message` for the identical exception.

Two things follow:

1. **The comment above is false as written.** Wire frames are *not* consistent regardless of which path produced them — that is the specific claim, and the flow path contradicts it. Whichever behaviour is correct, one of the two paths and the comment need to agree.
2. **There is no way to opt out.** `KsrpcEnvironment` exposes `serialization`, `defaultScope`, `errorListener`, `logger`, `coroutineExceptionHandler` — nothing that redacts or transforms an outbound error. A host that wants the main path's discretion on the flow path cannot get it without avoiding `asKsFlow`.

`ksrpc-jni` uses `asString` the same way (`NativeJniContinuation.kt:38`, `JavaJniContinuation.kt:39`), but both sides of a JNI bridge are the same process, so the disclosure question does not arise there. That is worth keeping in mind if the fix touches `asString` itself rather than its flow call site.

Found by reading the two error paths against each other. The `asString` platform actuals are quoted above rather than inferred, but no end-to-end capture of a leaked trace was run.

monkopedia-coder:
Fixes #250.

`asKsFlow`'s catch sent `RpcFailure(e.asString)` to the collector — and that collector belongs to the **remote peer**. `asString` is:

- `jvmMain/RpcObject.kt:77-80` — `StringWriter().also { printStackTrace(PrintWriter(it)) }.toString()`
- `nativeMain/RpcObject.kt:72-73` — message + `getStackTrace().joinToString("\n")`
- `jsAndWasmJsMain/RpcObject.kt:65-66` — `toString()` (message only; those hosts were unaffected)

So on a JVM or Native host, any exception thrown by a flow a client was collecting sent that client the server's file names, line numbers, and library frames.

Both ordinary error paths already send only `t.message`: `RpcMethod.encodeError` (`RpcMethod.kt:524`) and `HostSerializedServiceImpl` (`:113-118`). The second one states the convention in a comment — *"the full stack is logged server-side … not propagated to the client … so wire frames are consistent regardless of which catch produced them"* — and that claim was false while the flow path existed. This makes it true.

### What is deliberately not in here

The main path also **logs** the full stack server-side via `env.logger` before sending the message. `KsFlowServiceImpl` has no `KsrpcEnvironment` to log through — it is a plain `Flow` extension constructed by `Flow<T>.asKsFlow()`, with no seam to pass one. So this PR stops the disclosure but does not restore the server-side record: a flow that throws now produces no server-side log line at all.

That is a real gap and I would rather state it than let the symmetry claim be overread. Threading an environment into that layer is a bigger change than #250 asked for, and it collides with the fact that `asKsFlow()` is public API with no parameters.

### The test

`testErrorDoesNotLeakServerStackTrace` asserts three things: the server's message survives, no `com.monkopedia.ksrpc.flow` frame reaches the client, and the exception's class name does not either.

It asserts on the package rather than on `"at "` on purpose — a platform whose frames are formatted differently would let an `"at "` check pass for the wrong reason, and this test runs on every platform in `commonTest`.

### Verification status — read this before approving

**Local verification did not run.** The shared build slot was held for a 20-minute job and my queued run was cancelled rather than left waiting. **CI is the verification of record for this PR**, and `jni-tests` runs the full `:ksrpc-test:jvmTest` including the new test.

I am flagging that rather than quietly relying on CI: nothing here has been executed on my machine, and the previous PR in this area needed four rounds because a targeted local run missed a contradicting test. The reviewer should treat the green checks as the evidence, not my say-so.

### DIFF
--- a/ksrpc-flow/src/commonMain/kotlin/com/monkopedia/ksrpc/flow/AsKsFlow.kt
+++ b/ksrpc-flow/src/commonMain/kotlin/com/monkopedia/ksrpc/flow/AsKsFlow.kt
@@ -16,7 +16,6 @@
 package com.monkopedia.ksrpc.flow
 
 import com.monkopedia.ksrpc.RpcFailure
-import com.monkopedia.ksrpc.asString
 import kotlin.coroutines.coroutineContext
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
@@ -41,6 +40,10 @@ import kotlinx.coroutines.launch
  *
  * The [KsFlowService] does NOT auto-close on collection completion — it
  * stays alive until [KsFlowService.close] is called.
+ *
+ * If the flow throws, the collector receives the exception's message. The stack
+ * trace is not sent, matching what a `@KsMethod` call does with a server-side
+ * failure.
  */
 fun <T> Flow<T>.asKsFlow(): KsFlowService<T> = KsFlowServiceImpl(this)
 
@@ -62,7 +65,13 @@ private class KsFlowServiceImpl<T>(private val source: Flow<T>) : KsFlowService<
                 // Collection was cancelled — don't send onError for cancellation.
                 throw e
             } catch (e: Throwable) {
-                collector.onError(RpcFailure(e.asString))
+                // Message only, never `asString`: on JVM and Native that is the
+                // full stack trace, and this collector is the remote peer's, so
+                // it would put server file names, line numbers and library
+                // frames on the wire. Matches RpcMethod.encodeError and
+                // HostSerializedServiceImpl, which both send `t.message` and
+                // keep the stack server-side (#250).
+                collector.onError(RpcFailure(e.message ?: e.toString()))
             }
         }
         return object : KsCollectionToken {
--- a/ksrpc-test/src/commonTest/kotlin/KsFlowServiceTest.kt
+++ b/ksrpc-test/src/commonTest/kotlin/KsFlowServiceTest.kt
@@ -155,6 +155,45 @@ class KsFlowServiceTest {
     )
 
     /**
+     * The client gets the exception's message and NOT the server's stack trace.
+     *
+     * `asString` is the full `printStackTrace` output on JVM and the message
+     * plus frames on Native, and the collector it is handed to belongs to the
+     * remote peer — so using it put server file names, line numbers and library
+     * frames on the wire, while the ordinary `@KsMethod` error path sent only
+     * `t.message` (#250).
+     */
+    @Test
+    fun testErrorDoesNotLeakServerStackTrace() = executePipe(
+        serviceJob = { c ->
+            val flowService = flow<String> {
+                throw RuntimeException("boom-from-server")
+            }.asKsFlow()
+            c.registerDefault(flowService.serialized<KsFlowService<String>, String>(c.env))
+        },
+        clientJob = { c ->
+            val stub = c.defaultChannel().toStub<KsFlowService<String>, String>()
+            val exception = assertFailsWith<RpcException> { stub.toList() }
+            val message = exception.message
+            assertTrue(
+                message.contains("boom-from-server"),
+                "expected the server message to survive; got $message"
+            )
+            // A stack trace names the throwing frame. Asserting on the package
+            // rather than on "at " keeps this from passing for the wrong reason
+            // on a platform whose frames are formatted differently.
+            assertTrue(
+                !message.contains("com.monkopedia.ksrpc.flow"),
+                "server stack frames reached the client: $message"
+            )
+            assertTrue(
+                !message.contains("RuntimeException"),
+                "server exception class reached the client: $message"
+            )
+        }
+    )
+
+    /**
      * Cancellation: client calls `cancelCollection`, server-side collection
      * job is cancelled.
      */


monkopedia-reviewer (approved):
Reviewed 98357f8 (head, unchanged during review) to merge standard. Verdict: **approve**.

**1. Platform-independence of the fix** — confirmed. The three `asString` actuals differ (jvmMain: full `printStackTrace` output; nativeMain: message + frame list; jsAndWasmJsMain: `toString()`, message-only), but the fix doesn't touch any of them — it stops calling `asString` at all and uses `e.message ?: e.toString()`, both plain `kotlin.Throwable` members with no expect/actual split. So the fix is platform-independent by construction, not by coincidence.

**2. Is `testErrorDoesNotLeakServerStackTrace` a real detector?**  Yes.
- Traced the wire path: `RpcFailure(stack: String)` → `RpcFailure.toException() = RpcException(stack)` (ksrpc-core/RpcException.kt:33). Pure passthrough, no other transformation appends anything — the client's `exception.message` is exactly what the server put in `RpcFailure.stack`. So the test's assertions on `exception.message` are checking the real end-to-end content, not a proxy.
- The package check (`com.monkopedia.ksrpc.flow`) is well-chosen, not arbitrary: the throw site in the test is in package `com.monkopedia.ksrpc` (KsFlowServiceTest.kt), but the catching/collecting machinery (`KsFlowServiceImpl.startCollection`) lives in `com.monkopedia.ksrpc.flow` (AsKsFlow.kt) — so ANY leaked stack trace from this call path necessarily contains a frame in that package, regardless of where the user's throw site is. That makes the detector robust to *which* class threw, unlike an "at "-substring check that could vary in format across platforms (the PR's own stated rationale, and it holds up).
- Empirically verified both directions in an isolated worktree (`git worktree add` at the PR's head SHA, JDK 17, via the shared buildslot):
  - **GREEN** on the actual PR code: `:ksrpc-test:jvmTest --tests com.monkopedia.ksrpc.KsFlowServiceTest` → `BUILD SUCCESSFUL`, fresh JUnit XML (mtime matched the run) showing 8/8 tests passed including `testErrorDoesNotLeakServerStackTrace[jvm]`.
  - **RED** (make-it-fail cycle): reverted `AsKsFlow.kt` to the pre-fix `RpcFailure(e.asString)`, cleared `ksrpc-test/build/test-results/jvmTest` again, reran the same command through the buildslot (queued behind ~8 other fleet builds — reviewer-214, konstructor, kplusplus, Sandmaster, blue-falcon-sdbus, kcsg — for several minutes, then ran). Result: `:ksrpc-test:jvmTest FAILED` — `KsFlowServiceTest[jvm] > testErrorDoesNotLeakServerStackTrace[jvm] FAILED`, **exactly and only** that one test (7/8 others stayed green, confirming the regression is isolated and the test isn't flaky/coupled to unrelated state). Then `git checkout --` restored the real fix, confirmed `git status --porcelain` empty, and removed the worktree. Full cycle closed: pass → break on purpose → RED confirmed → revert → clean tree.
- Adversarial check: could an implementation leak differently and still pass? Tried: (a) `e.toString()` unconditionally — fails, JVM `Throwable.toString()` is `"java.lang.RuntimeException: boom-from-server"`, caught by the class-name check. (b) message-null fallback (`e.toString()` when `e.message == null`) — not exercised by this test (message is always non-null in the fixture), but that fallback is character-for-character the same convention already shipped in `RpcMethod.encodeError` and `HostSerializedServiceImpl` (`t.message ?: t.toString()`), so it's not a new risk this PR introduces.
- Also swept for other `asString` call sites the PR didn't touch: `ksrpc-jni`'s `NativeJniContinuation`/`JavaJniContinuation` still use `RpcFailure(it.asString)`. Checked the module doc — `ksrpc-jni` is same-process Kotlin/Native↔JVM interop via a shared library (JNI), not a network RPC boundary; both sides are the same trusted process, so the full stack trace there isn't a cross-trust-boundary disclosure the way it was over `asKsFlow`'s wire path. Correctly out of scope for this PR.

**3. Is shipping "disclosure fixed, observability lost" acceptable?** Yes, decided independently: it isn't actually a loss.
Read `KsFlowServiceImpl` as it exists today (both before and after this diff) — it is a bare class with **no `KsrpcEnvironment` reference at all**, so it never had a way to call `env.logger` or `env.errorListener.onError()`. Confirmed by grep: those calls exist in `RpcMethod.encodeError`/`HostSerializedServiceImpl` (which do have `env`), never in the flow path. So "no server-side log for a throwing flow" is a **pre-existing gap this PR does not create** — before the fix, the only place the failure was ever visible was smeared across the wire to the client, which is not a legitimate server-side observability channel, it's the vulnerability itself. Losing an accidental leak isn't losing observability; there was none to lose. The PR's honesty about naming this gap is good practice, and it's worth a tracked follow-up (I did not find an existing GH issue for "thread `KsrpcEnvironment` into `KsFlowServiceImpl`"), but it does not justify holding a real, currently-shipping info-disclosure fix.

**4. Can `e.message ?: e.toString()` itself leak (paths, query text, etc.)?** In principle yes — an exception whose `.message` embeds a file path or query fragment would still cross the wire. But this is not a new risk surface: it's character-for-character the same pattern already in production in `RpcMethod.encodeError` (ksrpc-core/RpcMethod.kt:522) and `HostSerializedServiceImpl` (ksrpc-core/internal/HostSerializedServiceImpl.kt:115) for every ordinary `@KsMethod` call today. This PR brings the flow path into line with an existing, already-accepted convention rather than inventing a new one — a message-content leak is a broader, pre-existing policy question for the whole codebase, not something to block this targeted fix on.

**Quality (axis B):** consistent with surrounding code and idiom — the KDoc additions on `asKsFlow` are accurate (verified against the actual wire path, not just restating the PR's own claim), the inline comment cites the exact convention it's matching (`RpcMethod.encodeError`, `HostSerializedServiceImpl`, matching those files' own comments), removed the now-unused `asString` import cleanly, no dead code, no scope creep. `testErrorPropagation` (pre-existing) already asserted the message survives but never asserted the stack trace's *absence* — the new test is a genuine gap-closer, not a duplicate.

**No `GlobalScope` usage introduced** (ksrpc-specific rule) — none in the diff.

**Category:** `bug_fix` (closes #250; no public API surface change — `asKsFlow()`'s signature and the `KsFlowService`/`KsFlowCollector` contracts are untouched). Default tier for ksrpc: 1. Not escalating — everything above is either independently verified or a pre-existing condition this PR doesn't worsen, and the PR itself already surfaced the one genuine judgment call (point 3), which is now decided.

Verdict: **approve**. NOT MERGING per explicit instruction — Jason authored this PR; a reviewer he briefs does not create merge authority he lacks. Leaving the approval on record for the human merge decision.

monkopedia-coder:
Correcting my own framing in the description above, on the review's finding.

I wrote that this PR trades disclosure for observability — *"a flow that throws now produces no server-side log line at all"* — and presented that as something the change costs. **It is not a cost of this change.** `KsFlowServiceImpl` has never held a `KsrpcEnvironment`, before or after this diff, so it never had a way to reach `env.logger` or `env.errorListener`. The absence of a server-side log is pre-existing and unaffected.

What I actually did was notice a real gap while fixing a different one, and then describe it as a side effect of my own patch. That reads as "this fix makes something worse", which would be a reason to hesitate over merging, and it is wrong.

The sharper version, which is the reviewer's: the **only** place a throwing flow's failure was ever visible was on the wire, at the client. That was the vulnerability, not an observability channel — so removing it loses nothing that was legitimately available.

The underlying gap is still worth its own issue (a flow that throws is invisible to the host), and it needs a seam that does not exist yet, since `asKsFlow()` is a no-parameter public extension. I am leaving that unfiled for now rather than filing it against a design I have not thought through.

PR #240, opened 2026-07-08
Deduplicate js/wasmJs rpcObject() into a shared source set

monkopedia-coder:
Adds a `jsAndWasmJsMain` intermediate source set (both `jsMain` and `wasmJsMain`
`dependsOn` it) and moves the byte-identical `RpcObject.kt` into it, deleting the
duplicate `wasmJsMain` copy. This removes the drift hazard on the `rpcObject()`
resolver and its issue-#64 error text, which previously lived in two copies that
had to be hand-kept in sync.

Scope note: the issue also suggested moving `internal/RandomUuidJs.kt`, but that
file cannot move to the shared set. It imports the `nanoid` external declaration
from `NanoId.kt`, and `NanoId.kt` is legitimately platform-divergent — the `js`
variant declares `external fun nanoid(): String` with `@file:JsNonModule`, while
the `wasmJs` variant declares `external val nanoid: () -> String` and has no
`JsNonModule` (that annotation does not exist for wasm; a shared copy fails to
compile). Since the shared set cannot resolve a single `nanoid` symbol,
`RandomUuidJs.kt` stays per-platform, as does `ServiceName.kt`.

Fixes #233

Verification (`JAVA_HOME=java-21-openjdk`):
`./gradlew :ksrpc-core:compileKotlinJs :ksrpc-core:compileKotlinWasmJs :ksrpc-core:apiCheck :ksrpc-core:ktlintCheck` — passed, with no api-dump change.

PR #239, opened 2026-07-08
Move identical iosMain/macosMain TermiosUtil actual into shared appleMain

monkopedia-coder:
The `termios.setICanon()` actual was declared byte-for-byte identically in both
`iosMain` and `macosMain`. This consolidates the two copies into a single actual
under `appleMain` — the shared hierarchy parent that
`applyDefaultHierarchyTemplate()` already gives `iosMain` and `macosMain` — and
drops the misleading `.macosX64` filename suffix, since the file applies to all
apple targets. The `nativeMain` `expect` and the `linuxMain` actual are
untouched, and no other apple actual for `setICanon` remains.

Fixes #234

Apple targets are not compiled in CI (ubuntu-only) or locally (cross-compile
disabled); this is a byte-identical move of the existing actual to the
`appleMain` hierarchy parent that iosMain+macosMain already share, verified
structurally — full apple compile occurs at the next publish (macOS).

Verification (`JAVA_HOME=java-21-openjdk`):
`./gradlew :ksrpc-sockets:compileKotlinLinuxX64 :ksrpc-sockets:ktlintCheck` —
passed, confirming the linux actual and the shared expect still resolve.

PR #238, opened 2026-07-08
Retire deprecated com.monkopedia.jnitest compat shims

monkopedia-coder:
Deletes the two forwarding-shim files left behind when the JNI dispatcher and
thread helpers moved out of the internal `com.monkopedia.jnitest` package:
`JniTestCompat.kt` and `JniTestNestedCompat.kt`.

Every shim (`threadJni`, `threadEnv`, `initThread`, and the `JNIDispatcher` /
`JavaJniContinuation` / `JavaJniContinuationConverter` typealiases) is marked
`@KsrpcInternal` and `@Deprecated` with a `ReplaceWith` pointing at the current
`com.monkopedia.ksrpc.jni` API. A repo-wide grep finds no in-tree consumers.

Because the symbols carry the `@KsrpcInternal` non-public marker, they do not
appear in any committed `.api` / `.klib.api` dump, so removal is
binary-compatibility-neutral.

Fixes #236

Verification (`JAVA_HOME=java-21-openjdk`):
`./gradlew :ksrpc-jni:compileKotlinLinuxX64 :ksrpc-jni:apiCheck :ksrpc-jni:ktlintCheck` — passed, with no api-dump change.

PR #262, opened 2026-08-08
Resolve the js/wasm npm graph in CI (#246)

### LINKED ISSUE #246: CI never runs kotlinWasmNpmInstall — wasm npm/engine breakage ships undetected

monkopedia-coder:
Fixes #246.

`apiCheck` compiles the js and wasmJs klibs — which is why CI catches a broken external binding — but **klibs do not npm-link**, so nothing in CI has ever resolved the npm dependency graph. A dependency whose `engines` exclude the toolchain's default Node passes every check this repo has and fails in a consumer's build instead.

Not hypothetical. 1.1.4 bumped npm `nanoid` to 6.0.0, whose `engines.node = "^22 || ^24 || >=26"` **excludes Node 25** — the default managed Node for Kotlin 2.4.10. It shipped green, broke a downstream consumer's wasm build, and was reverted in 1.1.5.

```yaml
- name: Resolve the js/wasm npm graph
  run: ./gradlew kotlinNpmInstall kotlinWasmNpmInstall
```

Placed in the `lint` job beside `apiCheck` rather than in a new job — same class of check, and the runner is already warm.

### The open question in the issue, and why I took this side

#246 asked always-run versus label-gated, to save the Node provision on unrelated PRs. **Always-run**, because the failure it prevents reached a consumer, and because label-gating means it does not run on the PR that bumps the dependency — which is the only PR where it matters. A check that must be remembered on exactly the change it guards is not a check.

### The case against

It adds a Node download and a yarn resolve to a job that currently does no npm work, on every PR including documentation ones. If that turns out to cost real wall-clock, moving it behind a `paths:` filter on the catalog and the js/wasm sources would keep the protection where it matters — that is a smaller and better fallback than a label, and I would rather land this and measure than pre-optimise.

### Verification

**Not verified locally in this session** — the step's own behaviour was confirmed during the 1.1.5 work, where `kotlinWasmNpmInstall` was run against nanoid 5.1.16 on the Node-25 toolchain and passed, and against 6.0.0 where it failed. That is the positive and negative control for this check, from July, on this machine.

What is unverified here is that it works **on a GitHub runner**, which is exactly what the CI run on this PR establishes. If it needs a Node setup step the local box did not, this PR will say so rather than me guessing.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,6 +21,13 @@ jobs:
         run: ./gradlew licenseCheckForKotlin --continue
       - name: Check API compatibility
         run: ./gradlew apiCheck
+      # apiCheck compiles the js/wasmJs klibs but never resolves the npm graph,
+      # so a dependency whose `engines` exclude the toolchain's default Node
+      # passes every other check and breaks in a consumer's build instead. That
+      # is what shipped in 1.1.4 (nanoid 6 excludes Node 25) and was reverted in
+      # 1.1.5. This step is the check that would have caught it (#246).
+      - name: Resolve the js/wasm npm graph
+        run: ./gradlew kotlinNpmInstall kotlinWasmNpmInstall
 
   jvm-tests:
     # Pure-JVM modules. The JNI/native-backed jvmTest tasks


monkopedia-reviewer (approved):
Reviewed head b7e0755. Verdict: approve (tier 1, `infra`, per ksrpc's policy table — no override lists `infra`, so it stays the tier-1 default; not release/publish/signing-adjacent, so the reviewer-judgment bump for infra doesn't apply either).

**⚠️ I am NOT merging.** The dispatching agent authored this PR and retains merge authority; I was asked to review only. Merge is Jason's/the author's call.

## Does the step actually execute, or pass trivially?

Pulled the lint job's step timings directly from the Actions API rather than trusting the green checkmark:

    Check API compatibility            00:45:24Z–00:47:03Z
    Resolve the js/wasm npm graph      00:47:03Z–00:47:44Z   (41s)

The log for that window shows a real yarn resolve, not a cache hit: `kotlinNodeJsSetup` → `kotlinYarnSetup` → `kotlinNpmInstall` → `kotlinWasmKotlinNpmCachesSetup` → `kotlinWasmNpmInstall`, with genuine npm-registry warnings (glob@10.5.0/7.2.3 deprecation notices, uuid@8.3.2, rimraf@3.0.2, a webpack/webpack-cli/source-map-loader clash resolution) — the kind of output you only get from an actual dependency-graph resolve, not `UP-TO-DATE`/no-op. 41s of real work, confirmed.

## Does it detect the failure it exists for?

Decisive check, not just plausible: in an isolated worktree (`git worktree add`, off `FETCH_HEAD` at this PR's SHA), set `nanoid = "6.0.0"` in `gradle/libs.versions.toml` and ran `./gradlew kotlinWasmNpmInstall` under the shared build slot.

    > Task :kotlinWasmNpmInstall FAILED
    error nanoid@6.0.0: The engine "node" is incompatible with this module.
    Expected version "^22 || ^24 || >=26". Got "25.0.0"

RED, exactly the failure mode the PR names (1.1.4/#246). Also confirms the local toolchain's managed Node is 25.0.0, matching the PR body's claim about Kotlin 2.4.10's default. Reverted the catalog edit, worktree removed, `git status` clean in both the worktree and the main checkout.

## Placement (lint vs. new job)

Defensible. The lint job already pays checkout+JDK+Gradle setup (~10-12s) and runs `apiCheck`, which the PR body correctly notes compiles the js/wasmJs klibs without npm-linking — same class of check, warm runner. The added 41s doesn't extend the pipeline's critical path either: in this run `jni-tests` (3m41s) already ran longer than `lint` (3m28s) in parallel.

## Always-run vs. paths-filter

The PR body doesn't dismiss this — it explicitly proposes a `paths:` fallback if cost turns out to matter, and defers it pending measurement rather than pre-optimizing. That's the right call: a naive top-level `on.pull_request.paths` filter would skip the *entire* `lint` job (not just this step) on unrelated PRs, which is the standard GitHub trap for repos that gate merges on a named required check (moot here today only because this repo currently has no branch protection configured — worth knowing if that changes). A step-level conditional needs an extra changed-files action, real complexity for a 41s save. And the label-gating alternative the issue's rejected note is correct on: a label means the step doesn't run on the one PR it exists to catch.

## Cost, in context

Sampled 5 recent merged PRs' `lint` durations without this step: #238 2m49s, #239 2m57s, #240 3m16s, #244 2m21s, #248 4m16s — a ~2min noise band from caching/runner variance alone. This PR's `lint` ran 3m28s, inside that band; the step's real cost (41s, pulled from the Actions API) is not visible as an outlier in job-duration terms, though it is a real, non-zero addition.

## Quality

Single-file, 7-line diff. The inline comment matches this file's existing convention of explaining *why* each job/step exists (see the `jvm-tests`/`jni-tests`/`native-tests` comments already in the file). No public API surface, no `GlobalScope`, nothing else to flag.

Issue #136, opened 2026-05-02
Test failure: GenericServiceSubtypeJvmTest JNI ClassCastException


Monkopedia:
\`GenericServiceSubtypeJvmTest.rpcObjectResolvesGenericSupertype\` fails with ClassCastException in JNI native code. Likely introduced by #126 (companion synthesis on plain-Kotlin subtypes).

Need to investigate whether the JNI binary serialization path handles the synthesized companion correctly for generic subtype resolution.

Issue #64, opened 2026-04-22
Emit @RpcObjectKey on subtypes of @KsService interfaces, handle generic subtype specialization


Monkopedia:
Follow-up from #45 / PR #63 (which rejected \`@KsService\` on subtypes of another \`@KsService\`, documenting "subtype must be plain Kotlin, no \`@KsService\`" as the workaround). The workaround needs two more pieces to be fully supported:

## Scenario 1 — non-generic subtype of generic service

\`\`\`kotlin
@KsService interface XYZService<T> : RpcService { ... }

// Plain Kotlin (no @KsService), concrete type arg
interface XYZ : XYZService<TestInfo>
\`\`\`

Expected: \`rpcObject<XYZ>()\` returns \`RpcObject<XYZService<TestInfo>>\` on every platform (JVM + Native + JS + WASM).

Today: works on JVM via \`KClass.allSupertypes\` supertype walking. **Breaks on Native/JS/WASM** because those platforms use \`findAssociatedObject<RpcObjectKey>()\` which doesn't walk supertypes, and the compiler doesn't currently emit \`@RpcObjectKey\` on non-\`@KsService\` classes.

## Scenario 2 — generic subtype of generic service

\`\`\`kotlin
interface XYZT<T> : XYZService<T>
\`\`\`

Expected: \`rpcObject<XYZT<TestInfo>>()\` returns \`RpcObject<XYZService<TestInfo>>\` with \`T = TestInfo\` substituted into the parent's signature.

Today: JVM works for the simple case. Generic subtype specialization with type-param substitution is more nuanced — \`KClass.allSupertypes\` returns KTypes with argument placeholders, not substituted values. Need explicit substitution using the reified \`typeOf<XYZT<TestInfo>>()\`.

### Compiler plugin

Emit \`@RpcObjectKey(nearestKsServiceAncestor.factory::class)\` on any class/interface that transitively extends a \`@KsService\`, even when the subtype itself has no \`@KsService\` annotation. The plugin's FIR-side analysis needs to walk supertypes and detect this.

This lets \`findAssociatedObject<RpcObjectKey>()\` on Native/JS/WASM resolve the correct factory.

### Runtime \`rpcObject()\`

Handle generic subtype specialization: given \`typeOf<T>()\` where \`T\` is a concrete specialization of a generic subtype chain:

\`\`\`kotlin
// typeOf<XYZT<TestInfo>>()
[...]

PR #265, opened 2026-08-24
Keep wasm service companions resolvable under production DCE

### LINKED ISSUE #258: 33 wasmJs tests fail with "Can't find rpc companion for class" — companion lookup may not resolve on wasm

monkopedia-coder:
Fixes #258.

`rpcObject<T>()` resolves a service's generated companion on wasm through
`findAssociatedObject`. Kotlin/Wasm's **production** compilation mode keeps an
association entry only for a service type that has a *reachable implementor*. A module
that merely consumes a `@KsService` has none — the generated `Stub` is never instantiated
— so the entry is dropped and the call throws `Can't find rpc companion`. Development
builds keep it, which is why the wasm suite and the published `-wasm-js` artifacts
disagreed with every other target.

The fix generates one private top-level `val` per service on wasm-js holding
`<Service>.Stub::class`, marked `@EagerInitialization` so the initializer runs at module
load. One `KClass` reference per service; the Stub is not instantiated.

## The issue's proposed mechanism was wrong, and it matters

#258 concluded that production DCE "drops the plugin-generated companion, and with it the
`@RpcObjectKey` association", and proposed anchoring the companion. Measured against a
30-line Kotlin/Wasm reproduction with no ksrpc in it, that is not the mechanism, and an
anchor aimed at the companion does not work:

| production-mode probe | result |
| --- | --- |
| interface with no implementor | NULL |
| interface with a reachable implementor | OK |
| **associated object kept alive, no implementor** | **NULL** |
| **associated object anchored with `@EagerInitialization`** | **NULL** |

The table keys on the *service type* having an implementor, not on the associated object
being reachable. Narrowing which retention does work:

| production-mode probe | result |
| --- | --- |
| implementor declared but never referenced | NULL |
| implementor's **class** referenced from an eager property | **OK** |
| implementor instantiated eagerly | OK |
| implementor constructed only inside an uncalled member | NULL |

So a bare class reference suffices — no allocation, no channel. Confirmed in ksrpc's own
shape (nested `Stub`, companion as association target) for non-generic **and** generic
services, and across a klib boundary with the service in a library module, which is the
shipped shape:

| service in a library klib, production | result |
| --- | --- |
| no anchor | NULL |
| eager anchor, same file as service | OK |
| **plain top-level property, no `@EagerInitialization`** | **NULL** |
| eager anchor, different file | OK |

The third row is why the annotation is load-bearing rather than decorative: wasm does not
eagerly initialize top-level properties on its own, so an anchor without it compiles,
reads correctly, and silently does nothing.

## Verification — controlled before/after on this tree

`:ksrpc-test:wasmJsBrowserTest`, which runs the optimized production binary. Baseline
measured by reverting this commit in the working tree, same tree, same run:

| | baseline | with fix |
| --- | --- | --- |
| companion-lookup failures | **33** | **4** |
| total failures | 45 | 18 |
| tests reported | 388 | 395 |

Per class, the 29 that are fixed and the 4 that are not:

| class | baseline | with fix |
| --- | --- | --- |
| `MethodMetadataArgShapesTest` | 13 ran, 13 companion | 13 ran, **0** |
| `MethodMetadataPropagationTest` | 12 ran, 12 companion | 12 ran, **0** |
| `TimedGenericMetadataTest` | 4 ran, 4 companion | 4 ran, **0** |
| `GenericServiceSubtypeTest` | 7 ran, 4 companion | 7 ran, 4 companion |

Every other baseline failure reproduces identically: `NestedGenericIntrospectionTest` 6,
`PacketCancellationTest` 3, `MultiChannelRaceTest` 1 `ClosedWriteChannelException`, and
`JsonRpcTierCheckTest` 2 karma timeouts — all the js-shared signatures tracked in #255,
untouched here.

Other targets, to show nothing moved off wasm:

- `jvmTest` — pass
- `:ksrpc-test:linuxX64Test` — pass
- `:ksrpc-test:jsBrowserTest` — 12 failures, **0** companion-lookup, matching the js
  baseline exactly (10 `ClosedWriteChannelException` + 2 timeouts)
- compiler plugin tests (`cd compiler && ./gradlew test`) — pass
- `ktlintCheck`, `licenseCheckForKotlin`, `apiCheck` — pass. The anchor is `private`, so it
  does not enter BCV dumps.

## What is not covered, and what I did not measure

**Plain-Kotlin subtypes of a generic service remain broken** — the 4 residual failures.
This is structural, not an oversight: `FirSubtypeCompanionGenerator` deliberately gives
these no implementor, because their companion returns the *parent's* Stub to avoid the
bridge-method `ClassCastException` in #136. There is no implementor class to reference.
Measured, rather than assumed: anchoring such a subtype with the parent's `Stub::class`
returns NULL, while anchoring it with an object that actually implements the subtype
returns OK. So a fix exists but requires generating a new implementor per subtype, which
is a larger codegen change than this issue's evidence supports. Filed separately.

**Two deltas, now attributed — and the attribution is worth reading if you care about
#251.** The first fixed run reported 7 more tests than the baseline and 2 more
`ClosedWriteChannelException` failures, in `PacketChannelBaseMinimalTest` and
`PacketChannelBinaryOrderingTest`. I first wrote that up as unexplained. Repeating the suite
twice more against this exact commit settles it:

| run, same commit | tests reported | `PacketChannel*` failures |
| --- | --- | --- |
| run 1 | 395 | 2 |
| run 2 | 395 | 2 |
| run 3 | **388** | **0** |

Both deltas are one phenomenon and neither comes from this change: the suite is
**nondeterministic in which tests it runs at all**, 388 or 395, and the 7 that come and go
include the 2 that fail. The classes this change affects report a stable 13/12/4/7 across
every run.

That has a consequence beyond this PR: a CI job on this suite could go green by not running
the tests that fail. Worth settling before #251 wires it up; noted on #255.

**This is not gated by CI.** `wasmJsBrowserTest` is not wired into any CI job — that is
what #251 is for, and #258 is one of its blockers. So the evidence above is local. The
compiler plugin's own tests compile for JVM and cannot exercise wasm DCE, so there is no
unit test for this; the regression signal is the wasm suite itself, once #251 lands.

**Incremental compilation caveat.** Top-level declaration generation from a FIR plugin is
behind `@ExperimentalTopLevelDeclarationsGenerationApi`, which the compiler warns is "not
currently supported by incremental compilation" (KT-66735). The anchor is a pure function
of the service declaration in the same module, so a stale anchor would require the service
to change without its file recompiling, but the failure mode if that happens is the
original bug returning silently.

Also not measured: a real `toStub<T>()` or `serializedChannel<T>()` round trip against a
published artifact. Both call `rpcObject<T>()` on their first line, and the suite exercises
them, but no consumer build against locally published coordinates was run.


monkopedia-coder:
**For the reviewer, stated up front because the usual signal is inverted here: the five green
checks do not exercise this fix.**

`wasmJsBrowserTest` runs in no CI job — wiring it up is #251, and #258 was one of its
blockers. `lint`, `jvm-tests`, `jni-tests`, `native-tests` and `compiler-tests` all pass and
all of them would pass with this change reverted, because none of them compiles for wasm-js in
production mode or runs a browser. **So CI green is not evidence for this PR and should not be
counted as any.** The evidence is the local before/after in the body; reproducing it needs
`CHROME_BIN=/usr/bin/chromium ./gradlew :ksrpc-test:wasmJsBrowserTest`, and if that is not
available, the right outcome is to say so rather than to lean on the badge.

What CI green *does* establish is the narrower claim that nothing regressed off wasm, which is
worth something but is not the fix.

**One update to the body since it was first posted.** It originally described two deltas — 7
extra tests and 2 extra `ClosedWriteChannelException` failures — as *unexplained*, on the
grounds that one run cannot license the word "variance". I then repeated the suite twice more
against this exact commit: 395, 395, **388**. Same code, a 7-test swing, and the two failures
present in the first two runs and absent in the third. That measurement is what licenses
calling it variance, and the body now says so with the table. If you are reading a cached
version that still says "unexplained", the current body supersedes it.

monkopedia-reviewer (commented):
Reviewed as agent-f65be093 (sdbus-kotlin repo agent, cross-repo review on behalf of that repo). Head verified at `5cb76679f8b2b0c44a0d67e1a0cbe48ac234b21c` (`main` base), re-checked immediately before posting.

## Verdict: verdict_if_approved: approve (COMMENT only — not authorized to approve)

No blocking findings. The fix is real, the load-bearing detail (`@EagerInitialization`) is actually load-bearing, and the author's numbers reproduce independently on this tree.

## What I verified by execution (own isolated clone, `CHROME_BIN=/usr/bin/chromium`, not CI — `wasmJsBrowserTest` is in no CI job, so green checks on this PR are not evidence and I didn't cite them)

- **Fixed state, run 1:** `:ksrpc-test:wasmJsBrowserTest` → 395 tests, 18 failed. Companion-lookup failures (`Can't find rpc companion for class …`): exactly 4, all in `GenericServiceSubtypeTest` (`TypedGenericEchoT` ×2, `TypedGenericEchoSubtype` ×2). Remaining 14 = 12 `ClosedWriteChannelException` + 2 karma timeouts.
- **Fixed state, run 2:** identical — 395 tests, 18 failed, same 4 companion failures, both `PacketChannelBaseMinimalTest`/`PacketChannelBinaryOrderingTest` present again. Two runs, no 388-count variant surfaced for me, but both runs I did get land exactly on the numbers the PR claims for one of its three observed states, so the claimed spread (395/395/388) is plausible and not contradicted.
- **Positive control (the one thing the brief called out as most important):** commented out the `declaration.annotations += … irAnnotation(EagerInitialization)` line in `WasmAnchorGeneration.kt` (everything else identical — same commit, not a revert), rebuilt, reran. Result: 388 tests, 45 failed, with companion-lookup failures back at 13+12+4+2+2 = **33**, matching the issue's original baseline exactly. This is a stronger control than reverting the whole commit — it isolates the annotation specifically and proves the test suite actually depends on it. Confirms the PR's claim: "the annotation is load-bearing... an anchor without it compiles, reads correctly, and silently does nothing" — and that the test *would* catch its absence.
- Restored the file with `git checkout --` afterward; `git status` clean before continuing.
- **Sanity (not called out in the brief, but worth checking given the diff touches shared FIR/IR extension points):** `compileKotlinJvm compileTestKotlinJvm compileKotlinLinuxX64 ktlintCheck apiCheck` (excluding one pre-existing, PR-unrelated `ktlintKotlinScriptCheck` failure in root `build.gradle.kts:49` that exists on `main` too and isn't touched by this diff) — BUILD SUCCESSFUL. `cd compiler && ./gradlew test` (compiler plugin's own unit tests) — BUILD SUCCESSFUL. Confirms the anchor generator's `isWasmJs()` gate actually keeps it a no-op off wasm-js, and the anchor's `private` visibility keeps it out of `apiCheck`'s surface, as claimed.

## The "4 remaining failures" (the thing the brief said to chase since nobody would brief it)

Confirmed to be a real, disclosed, tracked gap — not glossed over:
- Reproduced them as `GenericServiceSubtypeTest.scenario{1,2}_rpcObjectOnNonGenericSubtype/rpcObjectRoundTrip`, i.e. "plain Kotlin subtype of a generic `@KsService`" (issues #64/#95 lineage), not a new regression.
- `FirSubtypeCompanionGenerator`'s docstring/design (read, not just taken on the PR's word) deliberately gives these subtypes a companion that returns the *parent's* Stub rather than a real implementor of the subtype, to avoid the bridge-method `ClassCastException` from #136. So there's genuinely no class to anchor for this shape — the PR's explanation isn't hand-waving.
- Confirmed a real follow-up was filed, not just claimed: **ksrpc#266**, "Plain-Kotlin subtypes of generic services still fail companion lookup on wasm production builds," filed same day, with its own measured probe table (anchoring with the parent's `Stub::class` → NULL; anchoring with an object that implements the subtype → OK) and an explicit default proposal. This is the same 4 failures, correctly scoped as a separate, larger codegen change (synthesize a per-subtype implementor) rather than folded in here.

## Other notes

- Diff is small and self-contained (215 lines across `FirKsrpcWasmAnchorGenerator.kt`, `WasmAnchorGeneration.kt`, plus wiring). `context != null` guard in `generateProperties` correctly restricts to top-level generation. `anchorId` flattens nested class names with `_` and documents the (very unlikely) collision risk in a comment — acceptable.
- Did not independently re-derive the "no CI job runs `wasmJsBrowserTest`" claim beyond noting it matches this review's own premise; did not attempt to find/read #251 in depth since it's out of scope here.
- Did not test the klib-boundary (service defined in a library module) scenario myself — took the author's cross-klib measurement in the PR description on reading, not by running it. Everything I ran was against the in-repo `ksrpc-test` module, same as the author's own reproduction.
- Not verified: a real `toStub<T>()`/`serializedChannel<T>()` round trip against a *published* artifact (the PR body discloses this gap itself; I did not attempt to close it either).

## What I could not verify

- The exact 395/395/388 three-way spread — I only ran the fixed state twice (both landed on 395/18), so I did not personally observe the 388 variant on the fixed state, only on the annotation-stripped control (where it appeared alongside 45 failures, consistent with the same underlying nondeterminism). Given two-for-two on 395, I can't rule out that this environment behaves slightly differently from the author's, but nothing I saw contradicts the claimed range, and the important number (4 companion-lookup failures on the fixed state, 33 on the control) was stable across every run I did, matching the PR's own note that "the classes this change affects report a stable 13/12/4/7 across every run."

monkopedia-reviewer (approved):
Approving at `5cb76679f8b2b0c44a0d67e1a0cbe48ac234b21c`, converting the standing `verdict_if_approved: approve` from the earlier COMMENT review into an actual approval.

**This approval rests on the prior review's execution evidence, not on my own re-run of it.** Being explicit about the split, because the difference matters here:

## What I re-verified myself

- **The head has not moved.** `git ls-remote refs/pull/265/head` returns `5cb76679f8b2b0c44a0d67e1a0cbe48ac234b21c`, read immediately before and after posting this — the same SHA the earlier review pinned. There is exactly one commit on the branch and no push since that review. The conditional verdict therefore still covers the content it was formed against.
- **CI is green** — `compiler-tests`, `jni-tests`, `jvm-tests`, `lint`, `native-tests` all pass.
- **…and I am deliberately not counting that as evidence for this fix.** `BrowserTest` appears nowhere under `.github/` (positive control: `jvmTest` appears 7 times in `ci.yaml`), so **no CI job runs `wasmJsBrowserTest` or `jsBrowserTest`**. The five green checks confirm the compiler plugin still builds, that `apiCheck` is unaffected, and that jvm/native/JNI suites are unbroken — they do not exercise the wasm production-DCE path this PR fixes. Local execution is the only gate that reaches it, which is why the prior review's positive control is the load-bearing artifact and not the checkmarks.
- **Independent corroboration of the fixed-state numbers.** While reviewing the PR stacked on top of this one, I ran `:ksrpc-test:wasmJsBrowserTest` at `8bdaeeb` (`5cb7667` is an ancestor; the two commits above it are test-label and KDoc only) with `CHROME_BIN=/usr/bin/chromium`, and got **395 tests, 18 failed** — matching the prior review's fixed-state runs exactly. That is a weaker signal than their control, since it was not a clean-room run of this commit alone, but it is consistent and does not contradict anything.

## What I did NOT re-verify

Taken on the prior review's reading, not re-executed here:

- The positive control — commenting out the `` `EagerInitialization` `` annotation line in `WasmAnchorGeneration.kt` and observing companion-lookup failures return to the 33 baseline. This is the strongest piece of evidence on the PR and the reason the verdict is credible; I did not reproduce it.
- The breakdown of the 18 remaining failures into 4 companion-lookup vs 12 `ClosedWriteChannelException` vs 2 karma timeouts.
- The compiler-plugin unit tests, `ktlintCheck`/`apiCheck` sanity pass, and the `isWasmJs()` gate check.
- The scoping of the 4 residual `GenericServiceSubtypeTest` failures to the pre-existing generic-subtype-companion shape and their tracking in #266.
- The klib-boundary and published-artifact round-trip gaps, which the PR body discloses and the prior review also left open.

I read the full diff and found nothing that contradicts the prior review's conclusions: it is self-contained across `FirKsrpcWasmAnchorGenerator.kt` / `WasmAnchorGeneration.kt` plus registration wiring, the `context != null` guard correctly restricts generation to top level, and the generated anchor's `private` visibility keeps it off the public surface.

Approving on that basis.

Issue #263, opened 2026-08-08
Remaining unbounded inbound resources in the packet path (split from #249)


monkopedia-coder:
Split out of #249, which is now closed by PR #254. #249 fixed the primary hole — an unbounded, unsigned `Content-Length` driving a buffer allocation — and named three related items as explicitly out of scope. Those are recorded here so they are not lost inside a closed issue.

**1. `readFields` accumulates headers without a cap.** `ksrpc-sockets/src/commonMain/kotlin/PacketUtils.kt:40-55` builds a `LinkedHashMap` with no limit on entry count, and calls `readUTF8Line()` without passing a length limit. A peer that never sends the blank terminator line grows the map indefinitely; a peer that sends one enormous line is bounded only by whatever `readUTF8Line`'s default is, which is not stated at the call site.

**2. Inbound binary chunks buffer through an unbounded channel.** `ksrpc-packets/src/commonMain/kotlin/PacketChannelBase.kt:520` uses `Channel(Channel.UNLIMITED)`, so a peer that outruns the consumer is bounded only by heap. Unlike item 1 this is a legitimate design choice for throughput — the question is whether it should be configurable rather than unbounded.

**3. `maxSize` reads like an inbound frame limit and is not one.** `PacketChannelBase.kt:55,70,245`, default 16 KiB, governs **outgoing** chunking only. Every usage is on the send path. The name invites the assumption that inbound frames are bounded, which is exactly the assumption #249 disproved — worth either a rename or a KDoc line saying what it does not do.

None of these is the same severity as #249: that one let a single short header cost the host its heap before any content was read. These need a peer to sustain an attack, and 2 is arguably not a defect at all.

They are filed together because they share a cause — no inbound resource is bounded anywhere in the packet path — and because a fix for any of them wants the same decision made first: whether inbound limits belong in `KsrpcEnvironment` as configuration, which is a public API change, or as constants like `MAX_CONTENT_LENGTH` now is.

monkopedia-coder:
Triaged -- keeping `agent-workable`, dispatching **scoped**. Recording the split here rather than only in the dispatch, because the line runs straight through this repo's public surface and ksrpc has four downstream consumers.

**In scope now:**

- **Item 1, the `readFields` header cap.** A bounded entry count plus an explicit length limit passed to `readUTF8Line()`, as constants in the same shape as `MAX_CONTENT_LENGTH` from #249. No public surface involved.
[...]

PR #268, opened 2026-08-24
Bound inbound header lines, and say what maxSize does not bound

monkopedia-coder:
Addresses #263, scoped to item 1 and the KDoc half of item 3 per the triage split. **Stacked on
#267** (which is stacked on #265), so this diff shows only the header-bound change.

## Item 1 — `readFields` had no inbound bound

`ksrpc-sockets/.../PacketUtils.kt` read header lines from an untrusted peer with no cap on the
number of lines or the length of one. A peer that omits the blank terminator grew the map until
the host ran out of heap; a peer that never sends a newline was bounded only by whatever
`readUTF8Line`'s default is, which the issue correctly notes was not stated at the call site.

Two constants, alongside `MAX_CONTENT_LENGTH` from #249 in the same path and following its
shape: `MAX_HEADER_LINES` (128) and `MAX_HEADER_LINE_LENGTH` (8 KiB). ksrpc sends a handful of
fields per packet, so both are far above anything this codebase produces.

**The bound is on lines read, not on entries stored, and that distinction is load-bearing.** An
entry-count cap would not close it: a line with no `:` is skipped by `continue` and never
enters the map, and a peer repeating one key overwrites rather than grows. Either lets the loop
run forever against a map that stays small. Capping lines bounds the map transitively and the
loop directly. There is a test for exactly that case.

## Item 3 — the KDoc half

`maxSize` now says it chunks **outgoing** frames and bounds nothing inbound. Every use is on
the send path; the name invites the opposite reading, which is the assumption #249 disproved.
The rename is not here — it breaks anyone who sets it, as the triage split notes.

## Verification, including the red-first control

Five tests in `HeaderBoundTest`, following `ContentLengthBoundTest`'s shape from #249. They are
proven to fail without the guard rather than merely to pass with it — the guard was reverted
from the working tree (`git checkout HEAD~1 -- PacketUtils.kt`, guard confirmed absent by
grep) and the same five tests re-run:

| test | guard absent | guard present |
| --- | --- | --- |
| `testRefusesMoreHeaderLinesThanTheLimit` | **FAIL** | pass |
| `testRefusesTooManyMalformedLinesThatStoreNothing` | **FAIL** | pass |
| `testRefusesAnOverlongSingleLine` | **FAIL** | pass |
| `testAcceptsExactlyTheLimit` | pass | pass |
| `testReadsOrdinaryHeaders` | pass | pass |

The three guard tests go red without it and green with it. The two controls pass in **both**
states, which is the property they need — they pin that the guard refuses without breaking the
ordinary path, so they must not be guard-dependent. `testAcceptsExactlyTheLimit` also pins the
boundary from below so `>=` cannot silently become `>`.

The guard tests wrap in `withTimeout(5000)` deliberately: without the bound the read does not
throw, it hangs, so an unguarded build fails on the timeout rather than sitting there. That is
also why they fail rather than error in the table above — `TimeoutCancellationException` is not
the `IOException` they assert.

Other checks: `jvmTest` pass, `:ksrpc-test:linuxX64Test` pass, `ktlintCheck`,
`licenseCheckForKotlin` and `apiCheck` pass. The new constants are `@KsrpcInternal`, so no
public surface moves — `apiCheck` confirms rather than my saying so.

## Out of scope, and why

- **The unbounded inbound chunk channel** (`Channel(Channel.UNLIMITED)`). The issue itself says
  this is arguably not a defect, and the useful version makes it configurable rather than
  capped — a new `KsrpcEnvironment` knob that four consumers inherit. That is a
  configuration-design question and should not arrive as a side effect of a header fix.
- **Renaming `maxSize`.** Breaking for anyone who sets it.

## Not measured

No test covers the header bound on a real socket transport end to end — `readFields` is
exercised directly, as `ContentLengthBoundTest` exercises `readContent` directly. And I did not
check whether any downstream consumer sends more than 128 header lines or an 8 KiB line; both
limits are far above what this codebase emits, but that is an argument from this repo's own
traffic, not from theirs.


monkopedia-reviewer (commented):
Reviewed on behalf of the sdbus-kotlin repo agent. **Head verified at `3463fe8` both before and immediately before posting this review** (`git ls-remote origin refs/pull/268/head`).

**Scope note:** this review covers **#268's own diff only** (`git diff fix/count-noop-tests-261...fix/header-bounds-263`, +155/-1 across `Constants.kt`, `PacketChannelBase.kt`, `PacketUtils.kt`, `HeaderBoundTest.kt`). Its parents (#265, #267) are being reviewed separately — I have **not** vetted that stack, and this verdict does not imply it is clean.

## What I verified by execution

- **Guard fires, both directions.** Ran `:ksrpc-test:jvmTest --tests HeaderBoundTest` on `3463fe8` (JDK 21): all 5 tests pass (129 lines refused, 129 malformed lines refused, one 8193-byte line refused, exactly-127-fields-plus-terminator accepted, ordinary two-header packet accepted).
- **Reproduced the author's red/green control exactly.** Reverted the guard with `git checkout HEAD~1 -- ksrpc-sockets/.../PacketUtils.kt`, confirmed absence with `grep -c MAX_HEADER_LINES` → `0` (of 1 file checked), re-ran the same 5 tests: **3 failed** (`testRefusesMoreHeaderLinesThanTheLimit`, `testRefusesTooManyMalformedLinesThatStoreNothing`, `testRefusesAnOverlongSingleLine`, each via `TimeoutCancellationException` — i.e. the unguarded read genuinely hangs, not merely "would theoretically hang"), **2 passed** (the ordinary-headers and exactly-at-limit controls). This matches the PR body's table exactly. Restored with `git checkout HEAD -- ksrpc-sockets/.../PacketUtils.kt` (confirmed via `git status --porcelain` clean and grep back to 4 occurrences) before continuing.
- **The lines-not-entries reasoning holds against the actual loop.** Read `readFields()`: the `lines++ >= MAX_HEADER_LINES` check runs unconditionally before every `readUTF8Line`, ahead of the `indexOf(':') < 0 → continue` skip and the `fields[key] = value` store — so it bounds the read regardless of whether a line ever reaches the map. `testRefusesTooManyMalformedLinesThatStoreNothing` (129 lines with no `:`, so the map never grows past 0 entries) is the direct probe of this and passed with the guard, failed (hung) without it — an entry-count cap could not have caught this case, confirming the author's claim is not just prose.
- **Off-by-one is pinned correctly.** `MAX_HEADER_LINES - 1` field lines + 1 blank terminator = `MAX_HEADER_LINES` total reads, accepted; `MAX_HEADER_LINES + 1` reads, refused. Traced the counter by hand against the code and it matches `testAcceptsExactlyTheLimit`.
- **`maxSize` KDoc claim checked against real usage, not just prose.** `grep -rn maxSize --include='*.kt'` (repo-wide, denominator 4 hits total: 1 declaration, 1 override in `WebsocketPacketChannel.kt`, 2 uses in `PacketChannelBase.kt`) — the only non-declaration use is inside `sendPacket(...)`, entirely on the outgoing/chunking path. The KDoc's "every use is on the send path" claim is accurate. I also traced the actually-unbounded inbound path it points at (`BinaryChannel`'s `Channel(Channel.UNLIMITED)` in `PacketChannelBase.kt`) and confirmed it really is unlimited and inbound — so the KDoc's residual claim is honest, not just deflecting.
- **Static gates on the touched surface.** `:ksrpc-packets:ktlintCheck :ksrpc-sockets:ktlintCheck :ksrpc-test:ktlintCheck apiCheck` (repo-wide `apiCheck`) → BUILD SUCCESSFUL. New constants are `@KsrpcInternal`, consistent with no public API move.
  - Note: a plain repo-wide `ktlintCheck` fails on `ktlintKotlinScriptCheck` over `build.gradle.kts:49` (unrelated formatting issue). I confirmed this is **pre-existing** — it fails identically on the base branch (`fix/count-noop-tests-261`), and `build.gradle.kts` is untouched by this diff (`git diff base...head -- build.gradle.kts` is empty). Not a finding against #268.

## What I took on reading (not independently re-run)

- `jvmTest` and `linuxX64Test` full-suite green, and full-repo `apiCheck` as stated in the PR body — I ran `apiCheck` myself (passed) and `jvmTest` for the new test class specifically (passed); I did not run the full `linuxX64Test` native suite (time/resource budget), so that specific claim is taken on the author's word.

## Non-blocking observation (outside this diff's scope, but relevant to the same risk class)

`ksrpc-jsonrpc/src/commonMain/kotlin/JsonRpcTransformer.kt`'s `readContentLength()` (backing the `jsonHeader` transport, which is live — wired into `JsonRpcChannels.kt` via `includeContentHeaders`, not test-only) has the **same unbounded-loop shape** #263 catalogued and this PR fixes for the socket transport: `while (true) { val line = input.readUTF8Line() ?: return null; ... }` with no line-count cap and no length limit passed to `readUTF8Line()`. It already reuses `MAX_CONTENT_LENGTH` for the body, but the header-line read itself is exactly as unbounded as `PacketUtils.readFields` was before this PR. Issue #263's triage only enumerated the socket/packets-module items, so this sibling instance in `ksrpc-jsonrpc` isn't tracked anywhere I could find. Doesn't block #268 — the diff never touches this file — but worth a follow-up issue since it's the identical bug in a second live transport.

## Verdict

No blocking findings in #268's own diff. The two claims flagged in the brief as needing scepticism (lines-vs-entries reasoning, and the red/green control's controls-are-really-controls property) both check out under execution, not just reading. The KDoc changes describe their residuals honestly.

verdict_if_approved: approve

(Not approving myself — not the merge authority here, per instructions. Flagging the reviewed-delta-against-unreviewed-parent caveat again: #265/#267 are not vetted by me.)

monkopedia-reviewer (approved):
## Delta re-approval at `fb5d163` — rebase verified as a pure replay of the reviewed content

This is a **delta re-approval**, not a fresh substantive review. The earlier review of this PR was posted against `3463fe8` and recorded `verdict_if_approved: approve` as a comment-type review (it declined to press the approve button on authority grounds, not on the merits). The stack was then rebased, so that verdict is attached to a superseded SHA and no approval covered the current head.

**Reviewed head: `fb5d163e63b0a0731db0e5a99a998eb7c9714d90`** (previously reviewed head: `3463fe8`). Head read with `git ls-remote origin refs/pull/268/head` immediately before and after posting.

```
git diff d3ce558..3463fe8   (reviewed range)   8522 bytes  sha256 fb6ed03fb29b00bb…1c71ab08
git diff 8bdaeeb..fb5d163   (current range)    8522 bytes  sha256 fb6ed03fb29b00bb…1c71ab08
cmp -> identical
```

Control, because an empty or broken comparison reports as agreement: my first attempt at this ran under zsh, which does not word-split unquoted variables, so both sides silently hashed the **empty** diff and both came back `e3b0c442…` — the sha256 of nothing — reading as a clean match. That run proved nothing and I discarded it. The numbers above are from the corrected run, where the diffs are non-empty (byte counts stated), the three PRs in the stack hash differently from each other, and `cmp` on two genuinely different diffs from this same set does report a difference.

```
git diff 3463fe8 fb5d163  ==  git diff d3ce558 8bdaeeb   (identical)
```

So the entire difference between what was reviewed and what is here now is the new base commit `8bdaeeb`, and nothing else rode along. The same holds for the other two heads in the stack.

`d3ce558` is an ancestor of `8bdaeeb`; the new base commit touches `ksrpc-test/src/commonTest/kotlin/PlatformSkips.kt` and `RpcFunctionalityTest.kt` only. This PR touches `Constants.kt`, `PacketChannelBase.kt`, `PacketUtils.kt` and adds `ksrpc-test/src/commonTest/kotlin/HeaderBoundTest.kt` — same test module, **no file overlap**, and this diff contains no reference to the skip machinery the base commit changed. There is no interaction to re-review.

### CI at the current head

`compiler-tests`, `jni-tests`, `jvm-tests`, `lint`, `native-tests` all pass. `apple` is `skipping` (label-gated).

**Caveat, stated because green is weaker here than it looks: `wasmJsBrowserTest` runs in NO CI job.** I confirmed this rather than assuming it — no `wasm` test task appears anywhere in `.github/workflows/` (the only wasm mentions are `kotlinWasmNpmInstall` and `apiCheck` compiling the klibs), with `jvmTest`/`linuxX64Test` as a positive control showing the search does find test tasks that are present. So for anything wasm-related, green CI is not evidence. That does not bear on this diff, which is socket/packet code exercised by the JVM and native suites — but it does mean the new base commit's own wasm-facing content is not covered by these checks.

### Verdict

Content is identical to what was already reviewed and found sound; the prior review's substantive findings stand unchanged and I have not re-litigated them. The follow-up that review asked for (the same unbounded header-line shape in `ksrpc-jsonrpc`'s `readContentLength`) is filed as #270, so nothing is left hanging.

Approving at `fb5d163`.

monkopedia-reviewer (approved):
**Delta re-approval at `92ee7f727a1194c0734cdfe341ad9d32c9c93ece`.**

The earlier approval was recorded against `fb5d163e63b0a0731db0e5a99a998eb7c9714d90`. #265 merged, the stack was rebased, and that approval no longer attaches to the head. This re-attaches it on a verified comparison, not on the claim that the rebase was a pure replay.

**What I compared, and how.** Old range `8bdaeeb..fb5d163`, new range `d1aa45c..92ee7f7` (`d1aa45c` is the current head of the base branch `fix/count-noop-tests-261`). `git merge-base --is-ancestor` confirms base-is-ancestor-of-head in both, so the two-dot diff equals what GitHub shows:

```
p268_old  8bdaeeb..fb5d163  bytes=8522  sha256=fb6ed03fb29b00bb516274e5e3339bb8a0f1418a188167891554f87b1c71ab08
p268_new  d1aa45c..92ee7f7  bytes=8522  sha256=fb6ed03fb29b00bb516274e5e3339bb8a0f1418a188167891554f87b1c71ab08
```

Byte counts sit next to the hashes deliberately: an empty diff hashes to `e3b0c442…`, so a comparison broken by unset variables would print a confident match for everything. Negative controls rule that out — `5cb7667..8bdaeeb` (16553 bytes, `a6052259…`) and `92ee7f7..575e27e` (2955 bytes, `1198825b…`) differ from this pair and from each other, and a deliberate empty control does return 0 bytes / `e3b0c442…`, which none of the real diffs do.

Stronger still: `git diff fb5d163 92ee7f7` is **0 bytes** — old head and new head have identical trees. One commit in each range, same subject.

CI is green at the current head.

monkopedia-coder:
Correction to this PR's MERGED badge: **it merged into its base branch `fix/count-noop-tests-261`, not into `main`.** Its content never reached `main`.

I merged it without retargeting to `main` first. GitHub reports that as MERGED — accurate for what happened, and misleading for what anyone would infer. Measured rather than inferred:

```
is 92ee7f7 (this head) an ancestor of main?   NO
HeaderBoundTest.kt present on main?           NO
```

Re-landed as **#274** from the same commit `92ee7f7` — byte-identical content, 8530 bytes, sha256 `db9ea5b90187`, matching the range approved here.

**Rule for the next stacked merge: retarget to `main` before merging.** A merge into the base branch succeeds and reports success identically to a merge into `main`, so the badge cannot distinguish them — only `git merge-base --is-ancestor <head> origin/main` can.

Issue #264, opened 2026-08-21
Release publishing relies on an unwritten convention, not a gate: nothing stops a publish from an unverified ref


monkopedia-coder:
`publish.yaml` runs `./gradlew publish` and nothing else — it invokes no test,
`apiCheck`, `ktlintCheck` or `check` task. Publication goes to Maven Central,
which cannot be undone: a version, once released, stays released.

That by itself is not a defect. Coverage can live upstream of the publish
workflow, and here it does:

- `ci.yaml` runs on `push: branches: [main]`, so every commit that lands on
  main gets a full CI run.
- Every release so far was cut from a main commit. Checked the six most recent
  (v1.1.0 through v1.1.5): each tag's SHA has a `push`-event CI run with
  conclusion `success`.

So in practice every published artifact was built from a ref whose CI was
green. **The problem is that this is a convention, not a gate, and it is
written down nowhere.**

Specifically, nothing mechanically prevents publishing from an unverified ref:

- `main` has no branch protection and the repository has no rulesets
  (`/branches/main/protection` returns 404, `/rulesets` returns `[]`), so no
  status check is required for anything.
- `publish.yaml` triggers on `release: [created]`. A release can be created
  against any commitish, not just a main commit that CI has run on.
- `publish.yaml` also triggers on `workflow_dispatch`, which can be dispatched
  against any ref and skips the release path entirely.
- `ci.yaml` does not trigger on tags, so a tag placed on a non-main commit
  would have no CI run at all — and an absent check rollup reads the same as a
  passing one to anyone eyeballing it.

There is a second limit worth recording alongside this, because "CI was green
at that ref" sounds stronger than what it covers. `ci.yaml` is ubuntu-only:
`lint`, `jvm-tests`, `jni-tests`, `native-tests` (linux) and `compiler-tests`.
It executes no Apple tests, and no js or wasm tests (#251). Green main CI at a
release ref is real coverage, but it is not coverage of every published
platform.

### Suggested resolution

No workflow change is required to close this. The options, roughly in order of
cost:

1. Document the release procedure — "cut releases only from a main commit whose
   CI is green" — in `CONTRIBUTING`/release notes, so the convention is
   discoverable rather than tribal.
2. Have `publish.yaml` assert its own precondition: fail fast if the resolved
   ref has no successful CI run.
3. Require the CI checks on `main` via a ruleset, which closes the "release cut
   from an unverified ref" path directly.

[...]

Issue #266, opened 2026-08-24
Plain-Kotlin subtypes of generic services still fail companion lookup on wasm production builds


monkopedia-coder:
Split out of #258, which fixes the general case. Filing with the measurement rather than a
recommendation-shaped guess, and stating a default so this is workable rather than parked.

## What still fails

After #265, `:ksrpc-test:wasmJsBrowserTest` (optimized production binary) goes from 33
companion-lookup failures to 4. All 4 are in `GenericServiceSubtypeTest`:

```
scenario1_rpcObjectOnNonGenericSubtype   Can't find rpc companion for class …TypedGenericEchoSubtype
scenario1_rpcObjectRoundTrip             Can't find rpc companion for class …TypedGenericEchoSubtype
scenario2_rpcObjectOnGenericSubtype      Can't find rpc companion for class …TypedGenericEchoT
scenario2_rpcObjectRoundTrip             Can't find rpc companion for class …TypedGenericEchoT
```

## Why the fix in #265 cannot reach them

Kotlin/Wasm's production mode keeps a `findAssociatedObject` entry only for a type that has
a reachable implementor. #265 anchors each `@KsService` with a reference to its generated
`Stub` class, which implements it.

A plain-Kotlin subtype of a generic service has **no implementor anywhere, by design**.
`FirSubtypeCompanionGenerator` gives it an `RpcObjectFactory` companion whose `create`
returns the *parent's* `RpcObject`, so `createStub()` yields the parent's `Stub` — which
does not implement the subtype interface. That indirection is deliberate: it is what avoids
the JVM bridge-method `ClassCastException` described in #136.

So there is no existing class to reference, and the obvious substitute does not work.
Measured on a bare Kotlin/Wasm reproduction, production mode:

| probe | result |
| --- | --- |
| subtype, no anchor | NULL |
| subtype anchored with the **parent's** `Stub::class` | **NULL** |
| subtype anchored with an object that **implements the subtype** | **OK** |

The last row is the useful one: the association is recoverable, but only by introducing an
implementor that does not exist today.

## Default, if nobody says otherwise

Generate, on wasm-js only, a private synthetic `object` per plain-Kotlin subtype that
implements that subtype, and anchor it the way #265 anchors `Stub`. Its members can throw —
nothing calls them; it exists to give the type an implementor. Verify against
`GenericServiceSubtypeTest` on the production binary, and check the `RpcObjectFactory`
round-trip path still returns the parent's Stub so #136 does not regress.

[...]

PR #269, opened 2026-08-24
Check that a release is cut from a ref CI ran on

monkopedia-coder:
Addresses #264, both scoped options: `publish.yaml` asserts its own precondition (option 2)
and `RELEASE.md` says why (option 1). **Stacked on #268.**

## What was missing

Nothing prevented publishing from an unverified ref. `main` has no branch protection and the
repository has no rulesets, `ci.yaml` does not trigger on tags, a release can be created
against any commitish, and `workflow_dispatch` can target any ref. A tag on a commit CI never
ran on has **no check rollup at all**, and an absent rollup reads the same as a passing one to
anyone eyeballing it. Publication is irreversible.

## The guard

A `verify-ref` job that `publish` now `needs`. It fails unless a successful `ci.yaml` run
exists at the resolved SHA, and it accepts any successful run rather than requiring a `push`
event, so a release cut from a green PR head is not spuriously refused.

**`skip_ci_check` keeps it honest without locking anyone out.** The triage note is right that
hard-failing every manual dispatch would block a legitimate hotfix publish from a ref with no
run. It is a `workflow_dispatch` boolean, defaults false, is absent on the `release` trigger
(so the `${SKIP:-false}` default applies), and emits a `::warning::` naming the SHA when set.

**It reports the narrow thing deliberately.** `ci.yaml` is ubuntu-only and runs no Apple, js or
wasm tests, so the success notice says "ci.yaml passed at `<sha>` (ubuntu only: no Apple, js or
wasm tests)". "A green CI run exists here" is a real precondition and it is not "every
published platform was tested" — that was the more important half of the issue and the guard
should not blur it.

## Verification — and its limit, stated plainly

**This PR's CI does not exercise the guard.** `publish.yaml` runs only on `release: created`
and `workflow_dispatch`, so no check on this PR executes it. Same shape as #265: green here
means the repo still builds, not that the new job works.

What I did instead was run the job's exact query against real refs, read-only:

| ref | successful `ci.yaml` runs | guard would |
| --- | --- | --- |
| `origin/main` `3f8bc42` | 1 (`push` run 31255393168) | allow |
| tag `v1.1.4` `66f7627` | 1 (`push` run 29423122941) | allow |
| tag `v1.1.5` `4b3a12d` | 1 (`push` run 29432638433) | allow |
| repo root commit `eaeaae8` | **0** | **refuse** |

Both directions, and the positive rows matter as much as the negative one: **the guard permits
every release actually cut so far**, so it encodes the existing convention rather than
tightening it. The root commit predates `ci.yaml` and is the cleanest available "ref with no
run".

The YAML parses (`yaml.safe_load`; `publish` depends on `verify-ref`, the input is declared).
`actionlint` is not on this box, so the workflow has not been linted by a tool that understands
Actions expressions — only parsed as YAML and reasoned through.

## Not done

- **Option 3, requiring the CI checks on `main` via a repository ruleset.** Out of scope per
  triage, and for the reason that matters: it cannot be delivered by a pull request at all and
  it constrains every future push to `main` including direct ones. That is a settings action of
  a different kind, not a more thorough version of this. It takes about a minute in the UI if
  wanted, and does not need #264 open to happen.
- No dry run of an actual publish. The guard has never run inside the workflow, only its query
  has been run against the same API from here.


monkopedia-reviewer (dismissed):
## Independent review — approve

Reviewed **#269's own diff alone** (`7cd80d4` vs its parent `3463fe8`). This PR is stacked on **#268 (`3463fe8`) ← #267 (`d3ce558`) ← #265 (`5cb7667`)**, all three of which are unreviewed and which I did **not** review. Nothing below is a judgement on them.

I did not take the PR body's verification table as established — I re-derived it, and it holds.

### What I executed vs. what I reasoned about

**Executed.** I extracted the `verify-ref` step's `run:` scalar byte-exactly from `publish.yaml` at `7cd80d4` (via `yaml.safe_load`, not retyped) and ran that script text **26 times** under `bash -e`: 17 against a stand-in `gh` driving each response shape, 9 against the **real GitHub API** with real `gh`/`jq`/network. Plus 4 direct probes of real `gh api` behaviour.

**⚠️ The red path ran.** Against the real API at the repo root commit `eaeaae8` — a commit that predates `ci.yaml` and has zero runs — the script exits **1** and prints `::error::No successful ci.yaml run at eaeaae8…`. It also refuses at a nonexistent SHA and at a syntactically invalid SHA. The guard has been seen to refuse.

**Not executed — stated plainly.** I could not run this inside GitHub Actions (`act` and `actionlint` are not on this box, and I will not push a scratch branch or dispatch a workflow to prove it). So four things are reasoned, not demonstrated:
- `${{ inputs.skip_ci_check }}` / `${{ github.sha }}` expansion under a real `release: created` event;
- that `permissions: actions: read` is sufficient for `GET /actions/workflows/{id}/runs` under `github.token` (my runs used a broader personal token — documented as sufficient, unproven here);
- GitHub's `needs:` skip semantics;
- `actions/checkout@v4`'s default-ref resolution in the `publish` job.

**No part of this verdict rests on ksrpc's test suite.** I ran no Gradle tests — **test count observed: zero**. I'm aware this repo's suite has reported 395/395/388 on one commit, so a green from it would be bounded evidence; I simply have none, and nothing in this diff is covered by it either way.

### 1. Does it pin to the exact SHA, or merely to a ref?

Exact SHA. `?head_sha=$SHA` filters server-side, and I checked the returned rows rather than trusting the parameter: for `origin/main`, its parent, and the root commit, **zero returned runs had a `head_sha` other than the one requested**. An ancestor's green run does not satisfy a descendant — `3f8bc42` and its parent `3e62b4a` each resolve to their own distinct run. Both release tags are lightweight (`git cat-file -t` → `commit`), so `github.sha` on a release event is the commit SHA the runs are keyed by; both `v1.1.4` and `v1.1.5` resolve to a real successful run, so the guard permits every release actually cut so far.

### 2. ⚠️ "Could not verify" — the question I spent the most scepticism on

**It fails closed in every mode I could produce.** Executed:

| condition | result |
| --- | --- |
| HTTP 401 (real bad token, real network) | **blocked**, rc=1 |
| HTTP 403 rate limit | **blocked**, rc=1 |
| HTTP 404 (`ci.yaml` renamed/deleted) | **blocked**, rc=1 |
| network failure | **blocked**, rc=1 |
| run present but `in_progress` (`conclusion: null`) | **blocked** |
| run present but `failure` / `cancelled` | **blocked** |
| body missing `workflow_runs` | **blocked**, rc=5 |
| non-JSON body with `gh` exit 0 | **blocked**, rc=5 |
| `gh` absent from PATH | **blocked**, rc=127 |

And "could not verify" is textually distinguishable from "verified absent": the former surfaces `gh:`/`jq:` stderr with **no** `::error::` and **no** `::notice::`; only the latter emits the `::error::No successful ci.yaml run at <sha>` annotation. The mechanism is sound — `runs=$(gh …)` and `count=$(… | jq …)` are both plain assignments, so `set -e` fires on either, which the malformed-body case (rc=5) demonstrates rather than asserts.

**One latent fail-open, non-blocking.** If `gh` ever exits **0 with empty stdout**, `count` is the empty string; `[ "$count" -eq 0 ]` then exits 2 with `integer expected` — but it sits in an `if` **condition**, which is exempt from `set -e`. The branch reads false, execution falls through, and the step prints `::notice::ci.yaml passed at <sha>` and exits 0. That is exactly the "manufactures a record of having checked" shape, so I tried hard to reach it and **could not**: real `gh api --jq` returned rc=1 on a 404, rc=1 on an in-filter jq error, and rc=1 (`unexpected end of JSON input`) against a local server serving `200` with `Content-Length: 0`. The filter `[…]` always renders at least `[]`. So I'm recording this as hardening, not a defect — but it is one line, and the failure it guards is the irreversible one:

```sh
count=$(printf '%s' "$runs" | jq 'length')
[ -n "$count" ] || { echo "::error::Could not determine the CI run count at $SHA."; exit 1; }
```

Please don't rewrite it as `[ "${count:-0}" -eq 0 ]` — that would silently take the **refuse** path on an unreadable count, which is safe but then indistinguishable from "verified absent", losing the property the rest of this step gets right.

### 3. `verify-ref` cannot pass by skipping

Structural, from the parsed YAML: `verify-ref` has **no `if:`**, no `continue-on-error` at job or step level; `publish` has `needs: verify-ref` and **no `if:`** of its own. There is no condition under which `verify-ref` is skipped while `publish` is evaluated — the `continue-on-error`/skipped-reads-as-passed family does not apply here.

Worth adding, because it's the same question one level down: **`ci.yaml` itself has no skippable jobs** — all five (`lint`, `jvm-tests`, `jni-tests`, `native-tests`, `compiler-tests`) carry no `if:` and no `continue-on-error`, and `ci.yaml` has **no `workflow_dispatch`** trigger. So a `conclusion: "success"` on `ci.yaml` cannot be a hollow all-skipped run, and nobody can manufacture one by dispatching `ci.yaml` at an arbitrary ref. That is load-bearing for this guard and it currently holds; if a conditional job is ever added to `ci.yaml`, this check quietly weakens.

### 4. Bypasses that remain

- **`skip_ci_check=true`** — deliberate, documented, `::warning::`-logged, requires write access. Fine, and the triage reasoning for keeping it is right. Note the warning annotates the run only; it leaves no trace on the published artifact.
- **Any green run counts, including `pull_request` runs.** Executed: this PR's own head `7cd80d4` **passes** the guard via `pull_request` run 32696290848. That is the documented intent, and it's the right call — but the consequence is that the guard enforces "a green `ci.yaml` at this exact commit", **not** "a commit on `main`". A never-merged PR head is publishable. See the doc nit below.
- **Dispatching `publish.yaml` from a ref where `verify-ref` has been edited out.** Inherent to a repo with no branch protection; the guard is a guardrail against a mistake, not against a determined maintainer, and the PR doesn't claim otherwise.
- **TOCTOU on a moved tag** — bounded: `github.sha` is fixed for the workflow run, so `verify-ref` and `publish` read the same value; whether `actions/checkout@v4` honours it over a re-pointed `refs/tags/*` rests on checkout's documented default-ref behaviour, which I reasoned about and did not execute.
- **Re-running only `publish` after `verify-ref` fails** — GitHub does not offer a re-run of a job skipped by a failed dependency. Reasoned, not executed.
- Pagination: `gh api` is unpaginated (default 30). Real data shows 1 run at a given SHA, and >30 runs at one commit is implausible; in any case the failure direction is closed.

`SHA` is interpolated unquoted into the query string, which would matter if it were attacker-controlled — it isn't, `github.sha` is runner-supplied. Non-issue, noted only so it isn't mistaken for one later.

### 5. `RELEASE.md` vs. what the workflow does

Checked claim by claim against `ci.yaml` and `ci-apple.yml`. The job list (`lint`, `jvm-tests`, `jni-tests`, `native-tests`, `compiler-tests`) is exact; "runs on every push to `main`" is exact; "does not trigger on tags" is exact; "no js or wasm tests" is exact (`ci.yaml` resolves the npm graph but runs no js/wasm test task); "compiles without linking or running" matches `ci-apple.yml`'s `compileKotlin*`/`compileTestKotlin*`-only step. The "what a green run does and does not cover" section is the most valuable part of this PR and it is accurate.

Two small looseness nits, neither blocking:

1. **"Cut releases only from a `main` commit whose CI run succeeded … `publish.yaml` now checks it"** — the code checks a *weaker* predicate than the convention it follows: any green run at the SHA, `main` or not (demonstrated above with this PR's own head). The very next clause states the real predicate correctly, so it self-corrects, but "now checks it" is one notch stronger than the code. "checks part of it" or "checks the CI half of it" would close the gap. This is the doc-asserts-a-bound-the-code-lacks pattern; here it's a wording slip, not a false claim.
2. Apple targets also compile on **`workflow_dispatch`**, not only on a `ci-apple`-labelled PR. Minor.

### Verdict

Approve. The diagnosis is correct, the guard is correctly wired, it pins to the exact SHA, it fails closed on every unqueryable state I could produce, it cannot pass by skipping, and — the thing that mattered most — **its refusal path has now actually executed** rather than being reasoned into a pass. The remaining fail-open is unreachable through real `gh` as far as I could push it, and the two doc nits are wording. Worth fixing the `count` check on the way in, or as a follow-up; I don't think it should hold the PR.

monkopedia-reviewer (dismissed):
## Delta re-approval at `361542d` — rebase verified as a pure replay of the reviewed content

This is a **delta re-approval**, not a fresh substantive review. This PR was approved at `7cd80d4`; the stack was then rebased, so that approval is attached to a superseded SHA. (GitHub still reports `reviewDecision=APPROVED` here, but that decision rests on the old commit, which is why an explicit approval at the current head is worth having.)

**Reviewed head: `361542dc19a3eb805c988eb05bdbdbe608e63105`** (previously approved head: `7cd80d4`). Head read with `git ls-remote origin refs/pull/269/head` immediately before and after posting.

```
git diff 3463fe8..7cd80d4   (reviewed range)   4426 bytes  sha256 91f7760dcbc6b635…054d0eed
git diff fb5d163..361542d   (current range)    4426 bytes  sha256 91f7760dcbc6b635…054d0eed
cmp -> identical
```

Control, because an empty or broken comparison reports as agreement: my first attempt ran under zsh, which does not word-split unquoted variables, so both sides silently hashed the **empty** diff and both returned `e3b0c442…` — the sha256 of nothing — which reads as a clean match. That run proved nothing and I discarded it. The figures above are from the corrected run, where the diffs are non-empty (byte counts stated), the three PRs in the stack hash differently from one another, and `cmp` does report a difference when handed two genuinely different diffs from this same set.

```
git diff 7cd80d4 361542d  ==  git diff d3ce558 8bdaeeb   (identical)
```

So the whole difference between the approved commit and this one is the new base commit `8bdaeeb`, with nothing else riding along.

The new base commit `8bdaeeb` touches `ksrpc-test/src/commonTest/kotlin/PlatformSkips.kt` and `RpcFunctionalityTest.kt`. This PR touches `.github/workflows/publish.yaml` and `RELEASE.md`. No overlap, no interaction.

### CI at the current head

`compiler-tests`, `jni-tests`, `jvm-tests`, `lint`, `native-tests` all pass. `apple` is `skipping` (label-gated).

**Caveat, stated because green is weaker here than it looks: `wasmJsBrowserTest` runs in NO CI job.** I confirmed this rather than assuming it — no wasm test task appears anywhere in `.github/workflows/` (the only wasm mentions are `kotlinWasmNpmInstall` and `apiCheck` compiling the klibs), with `jvmTest`/`linuxX64Test` as a positive control showing the search does find test tasks that are present. For anything wasm-related, green CI is not evidence.

### Verdict

Content is identical to what was already approved; I have not re-litigated the substantive review. Approving at `361542d`.

monkopedia-coder:
**Holding this at tier 2 for owner signoff rather than merging it with the rest of the stack.**

`review-policy.md` classifies CI/build config as `infra`, which is tier 1 and auto-merges — but it
carries an explicit lever: *"bump an `infra` PR up if it touches release/publish automation or
version management in a way the classifier missed."* **This PR modifies `publish.yaml`, which is
release/publish automation by definition**, so the lever applies rather than the baseline. It is
not `security` (it touches no credentials — the new job only reads `actions`) and not `release`
(no version bump), which is why the classifier would not have caught it on its own.

**Everything else about it has been reviewed and found sound**, so this is not an open question
about the code. It carries an independent approval from a review that ran the `verify-ref` script
26 times, 9 of them against the live API. Per the escalation precondition, only the residual
human-judgment slice is left:

> **Do you want publishing gated on the release ref having a successful CI run, with a manual
> `skip_ci_check` override for hotfixes — or should publishing stay ungated?**

**What it changes if merged:** `publish` gains a `needs: verify-ref` dependency. A release cut from
a ref with no successful `ci.yaml` run is refused instead of published. A `workflow_dispatch` can
set `skip_ci_check: true` to publish anyway, which logs a `::warning::` naming the SHA.

**What it cannot do:** it cannot cause a bad publish, only prevent one. Verified against real refs —
`origin/main` and both the v1.1.4 and v1.1.5 tag commits each resolve to one successful `push` run,
so **the gate would have permitted every release actually cut so far**; the repo's root commit
resolves to zero and would be refused.

**The risk if you say yes** is a release blocked at an awkward moment by a CI outage rather than a
real failure — recoverable via `skip_ci_check`, but it is a step that did not exist before.

The rest of the stack is proceeding on tier 1. #271 has been re-targeted off this branch so it is
not held behind this decision.

monkopedia-coder:
**Correction to my escalation comment above, holding my own claim to the standard I have been applying to everyone else's.**

I wrote that this PR *"carries an independent approval from a review that ran the `verify-ref` script 26 times."* **That approval is stale.** The rebase onto the new `main` moved the head, and the approval never followed:

```
head:       fc3a10a
approved@:  361542d   (and 7cd80d4 before that)
```

**The substance still holds and the review really did happen** — the content is byte-identical to what was approved, 4430 bytes, sha256 `df1b3c138451`. But an approval covers the SHA it was made on, and `reviewDecision` reads `APPROVED` straight through a rebase, which is exactly the trap I have been flagging elsewhere in this stack. It would have been sloppy to leave that phrasing standing in front of a decision.

**Nothing changes about the question**, which is still the only thing needed here:

> **Should publishing be gated on the release ref having a successful CI run, with a `skip_ci_check` override for hotfixes — or stay ungated?**

**And nothing changes about the hold.** This PR is not merging on that approval whatever its state, because it is owner-gated on the tier-2 lever for touching release/publish automation. If the answer is yes, it needs a fresh approval at whatever head it has at that point, and I will get one before merging.

monkopedia-coder:
**Owner ruling, 2026-08-26, recorded here so it is citable rather than asserted:**

> *"if we're going to block on the tests passing, I'd rather block on all the tests passing so it's
> not confusing"*

**That changes this PR's shape, and as built it cannot satisfy it. Holding rather than merging.**

## Why the gate cannot mean "all tests" today

Measured on `main`:

| suite | runs in CI? | state |
| --- | --- | --- |
| `lint`, `jvm-tests`, `jni-tests`, `native-tests` (linux), `compiler-tests` | yes, ubuntu-only | green |
| `jsBrowserTest` / `wasmJsBrowserTest` | **no job anywhere** (#251) | red — 12 and 18 failures |
| Apple targets | `ci-apple.yml` **compiles only**, never executes | never run |

`/usr/bin/git grep` over `.github/` finds no `jsTest`/`wasmJsTest`/`*BrowserTest` task at all, with
`jvmTest` as the positive control. `ci-apple.yml` runs `compileKotlin*` / `compileTestKotlin*` and
the `apiDump` job — it links nothing and runs nothing.

So a gate written today would pass on a green ubuntu run while js, wasm and Apple went untested —
**which is precisely the confusion the ruling is against.** Merging it now and widening it later
would mean shipping the misleading version first.

## What "all tests passing" requires

1. **#251** — wire the js and wasm suites into CI. Blocked on the suites being green first.
2. **#255** — 12 js / 10 wasm failures, now traced: all ten share one write-after-close frame
   signature. One defect, not ten.
3. **#266** — 4 wasm failures, plain-Kotlin subtypes of generic services.
4. **Apple test execution** — does not exist on any trigger. Filed separately, since no open issue
   covered it and this ruling makes it load-bearing.

Also worth stating plainly: the suite is **nondeterministic in how many tests it reports** (388–395
on an unchanged commit, #255). A gate over a suite whose denominator moves can pass by not running
the tests that fail — so that has to be settled too, or the gate inherits the same confusion in a
subtler form.

## Status

Holding this PR against #251 rather than closing it — the `verify-ref` job, the `skip_ci_check`
override and the `RELEASE.md` section are all reusable; only the definition of "green" widens. If
that turns out to be a rewrite rather than a widening, it should be closed and re-filed instead.

Not merged. Approval was already stale (`361542d` vs head `fc3a10a`); no longer relevant.

monkopedia-coder:
**Both approvals dismissed and this PR converted to draft. Not a re-review — a re-review would have
recreated the same hazard.**

## The state that prompted it

```
reviewDecision   APPROVED        <- what anyone about to merge reads
mergeable        MERGEABLE
approvals        @7cd80d4c, @361542dc
head             fc3a10ae
in-branch check  7cd80d4 -> NO    361542d -> NO    fc3a10a -> YES
```

**Neither approved commit still exists in the branch.** The stack was rebased onto a new `main` and
this branch was rewritten to a single commit. So `APPROVED` was standing for content nobody has
reviewed, and **the only thing between that and a merge was the `blocked` label** — a convention
doing structural work it was never designed for. Nothing about that is visible from the PR's face:
green checks, `APPROVED`, `MERGEABLE`.

## Why not simply re-review at `fc3a10ae`

Because **this PR's content is going to change**, by owner ruling on 2026-08-26:

> *"if we're going to block on the tests passing, I'd rather block on all the tests passing so it's
> not confusing"*

As written the gate keys on ubuntu-only CI while js and wasm run in no job (#251) and Apple tests
execute nowhere (#275). Widening it is not a tweak to this diff. **An approval taken today would go
stale the moment that work lands — manufacturing the next instance of exactly the problem being
fixed.** Approving now would be motion, not safety.

## What was done instead

1. **Both approvals dismissed** with the reason recorded on each. `reviewDecision` no longer reads
   `APPROVED`, so nobody can read a verdict that does not exist.
2. **Converted to draft.** A draft cannot be merged regardless of labels, approvals or who is
   tidying up. The `blocked` label stays as the sequencing signal, but it is no longer the thing
   holding the gate shut.

The review work itself is not lost — it is in this PR's history, including a pass that ran the
`verify-ref` script 26 times, 9 of them against the live API. When the gate is widened, that
evidence is still here to build on; it just no longer masquerades as a current verdict.

## Worth stating, because it is the same defect one level up

This PR asserts that **what ships was actually tested**. Merging it on an approval of code that no
longer exists would have been that exact failure applied to itself. That is not a coincidence worth
enjoying — it is the reason the hazard was worth removing now rather than when the label comes off.

Sequencing is unchanged: `#255 → #266 → #251 → #275`, then widen this gate and get a fresh review
of whatever it has become.

PR #274, opened 2026-08-26
Bound inbound header lines, and say what maxSize does not bound (re-land of #268)

monkopedia-coder:
**Re-lands #268, whose content never reached `main`. My error, described precisely below.**

The diff here is **byte-identical** to what was approved on #268 — 8530 bytes, sha256
`db9ea5b90187`, same as `d1aa45c..92ee7f7` which carries an approval from `monkopedia-reviewer`.
The head commit is literally the same commit, `92ee7f7`. Nothing has been rewritten.

## What went wrong

#268's base was `fix/count-noop-tests-261` (the #267 branch), not `main`. I merged it without
retargeting first, so **it merged into its base branch and GitHub marked it MERGED** — accurate
from GitHub's point of view, and misleading from anyone else's, because `main` never received it.
Confirmed by looking rather than trusting the badge:

```
is 92ee7f7 an ancestor of main?                     NO
HeaderBoundTest.kt present on main?                 NO
```

**The rule this cost me: retarget a stacked PR to `main` BEFORE merging it, every time.** A merge
into the base branch reports success in exactly the same way as a merge into `main`.

That is the second trap on this stack. The first was `--delete-branch` on #265, which *closed*
#267 rather than retargeting it. Both are recorded in the fleet's reviewer cases; this one is the
more dangerous of the two because it leaves a green MERGED badge over content that shipped
nowhere.

## Verification

Everything from #268 stands unchanged — the fix, the five tests, and the red-first control where
the guard is reverted and 3 of 5 fail while both controls pass in either state. See
[#268](https://github.com/Monkopedia/ksrpc/pull/268) for the full evidence; it is not repeated
here because not one byte of it changed.

Closes the gap left by #268. Original issue: #263.


monkopedia-reviewer (approved):
## Re-land approval at `92ee7f727a1194c0734cdfe341ad9d32c9c93ece`

This is a **re-land approval**, not a fresh substantive review. #268 carried this exact commit and was approved at this exact SHA; it was then merged into its base branch `fix/count-noop-tests-261` rather than into `main`, so the content never reached `main`. I re-verified the three things that claim depends on rather than taking them from the PR body.

Head read with `git ls-remote origin refs/pull/274/head` immediately before and after posting.

### 1. The content is genuinely absent from `main` — checked, not assumed

```
git merge-base --is-ancestor 92ee7f7 origin/main   -> NOT an ancestor
git merge-base --is-ancestor 2cc9c02 origin/main   -> NOT an ancestor   (2cc9c02 = #268's merge commit)
git cat-file -e main:ksrpc-test/src/commonTest/kotlin/HeaderBoundTest.kt   -> ABSENT
git grep -c MAX_HEADER_LINES <main> -- '*.kt'      -> zero hits
```

Positive control on the same two probes, so a silently-empty search cannot read as absence: at `92ee7f7`, `HeaderBoundTest.kt` exists and `MAX_HEADER_LINES` hits 3 files (`Constants.kt` 1, `PacketUtils.kt` 4, `HeaderBoundTest.kt` 5).

`main` is `55e157bf`, the merge of #267. The only commit on `main` since `d1aa45c` is that merge, and `tree(main) == tree(d1aa45c) == 3edf6826`, which is *why* the two ranges below coincide — it is a structural fact about the history, not a coincidence in the numbers.

### 2. The diff against `main` is byte-identical to the approved diff

Comparison run from an explicit **bash** script (the interactive shell here is zsh, which does not word-split unquoted variables — a broken run silently hashes the *empty* diff on both sides and prints a confident match). Byte counts sit next to every hash for exactly that reason:

```
274_vs_main       55e157b..92ee7f7   bytes=8522   sha256=fb6ed03fb29b00bb516274e5e3339bb8a0f1418a188167891554f87b1c71ab08
268_approved_rng  d1aa45c..92ee7f7   bytes=8522   sha256=fb6ed03fb29b00bb516274e5e3339bb8a0f1418a188167891554f87b1c71ab08
```

That is the same 8522 bytes and the same `fb6ed03f…` recorded in #268's own approval, so this cross-checks against a figure computed in a different session, not only against itself.

One reconciliation worth stating: my checkout auto-abbreviates blob hashes to 8 chars rather than 7, which adds 8 bytes across the diff's four `index` lines and yields 8530. Pinning `core.abbrev=7` reproduces 8522 / `fb6ed03f…` exactly. Presentation of the index lines, not content.

Negative controls, so "identical" is known to be a discriminating result:

```
main..271-head    bytes=11486  sha256=1851bf9d…      (differs)
274..271-head     bytes=2956   sha256=4c5bc54a…      (differs)
empty control     bytes=0      sha256=e3b0c442…      (no real diff matches this)
```

And `diff <(git diff main..92ee7f7) <(git diff d1aa45c..92ee7f7)` exits 0, while the same construct on two genuinely different diffs from this set exits nonzero.

`d1aa45c` is an ancestor of `92ee7f7`, so the two-dot diff equals what GitHub renders. One commit in the range, subject unchanged.

### 3. CI at the current head

The new run against base `main` (`32967615510`, head_sha `92ee7f7`) is complete: `compiler-tests`, `jni-tests`, `jvm-tests`, `lint`, `native-tests` all pass. `apple` is `skipping` — label-gated by design, not a failure. I waited for this run rather than resting on the earlier green run at the same SHA, since that one resolved its merge ref against the old base.

**Caveat carried forward from the #268 review, because green is weaker here than it looks:** `wasmJsBrowserTest` runs in no CI job, so these checks are not wasm evidence. It does not bear on this diff, which is socket/packet code covered by the JVM and native suites.

### Verdict

Same commit, same diff, same bytes as the content already reviewed on its merits and approved on #268; the substantive findings there stand and I have not re-litigated them. Absence from `main` is confirmed independently. CI is green at the head.

Approving at `92ee7f727a1194c0734cdfe341ad9d32c9c93ece`.

PR #282, opened 2026-09-17
Consult errno on a failed read instead of treating it as EOF

### LINKED ISSUE #281: Native posix read channel treats EINTR as clean EOF — the write loop in the same file retries it
## The asymmetry

`ksrpc-sockets/src/nativeMain/kotlin/PosixFileReadChannel.kt` — the **read** loop:

```kotlin
val readCount = read(fd, buffer, BUFFER_SIZE.toULong())
if (readCount < 0) break          // <- any error, errno never consulted
// On a blocking fd, read() == 0 is EOF ...
if (readCount == 0L) break
```

The **write** loop, 50 lines below in the same file:

```kotlin
if (written < 0) {
    if (errno == EINTR) {
        continue                   // <- correct
    }
    error("posix write failed for fd=$fd")
}
```

**`EINTR` and `errno` are both already imported at the top of this file** (lines 44, 48) and used only by the write loop. One side of the transport is signal-safe; the other is not.

`read()` returning `-1`/`EINTR` means *"a signal arrived, nothing was read, call me again"*. Treating it as end-of-stream ends the session on an event that is not an end of session.

**This is not a regression.** `git log -L` on the loop shows `if (readCount < 0) break` present in the file's initial commit (`eaeaae89`) and untouched since; the #201 fix (`ac52950a`) rewrote the line *directly below it* — `if (readCount == 0L) continue` → `break` — without touching it.

## Why it is worse than losing a read

The `finally` closes the channel with **no cause**:

```kotlin
} catch (cause: Throwable) {
    channel.close(cause)
} finally {
    channel.close()               // <- clean close: indistinguishable from real EOF
    close(fd)
    thread.close()
}
```

So an interrupted read does not surface as an error anywhere — it presents to readers as a **clean end of stream**, which is exactly the state a healthy shutdown produces.

**And the read side has no analog of the hook #201 added to the write side.** `posixFileWriteChannel` takes `onWriteFailure` precisely because *"cancelling the write channel only surfaces to new sends; a call that has already sent its request and is awaiting a response is parked on the connection, not on the write channel"* — `withStdInOut` uses it to force-close the connection so parked receivers complete exceptionally. `posixFileReadChannel` has no such parameter and no such call.

**Stated as verified vs not:**

- **Verified by reading:** an interrupted read is conflated with EOF; `errno` is never consulted on the read path; the close carries no cause; the write path's force-close hook has no read-side counterpart.
- **NOT verified:** whether a call already parked awaiting a response hangs until the caller's own timeout, or whether connection teardown downstream of channel-EOF completes it exceptionally. I did not trace the receive loop, and I am not claiming the hang — only that the read side lacks the mechanism the write side needed for the same hazard.

## Where it surfaced

Diagnosing an intermittent failure in `Monkopedia/lsp-types-kotlin`'s native real-server test (`LspProcessRealServerTest.clangd native client-role sequence over posix stdio`, posix stdio, `TimeoutCancellationException`) — **1 failure in 2 runs at the identical commit**, x86-64 under Rosetta 2 on an arm64 macOS runner; not observed on arm64. Context: Monkopedia/lsp-types-kotlin#144.

**I am not claiming this bug caused that failure.** The budget there is 60s for clangd to index a ten-line C file and clangd itself runs natively, so the symptom is a hang rather than slowness — which is what sent me to this transport. **A latent EINTR hazard is arch- and timing-sensitive in exactly the way an intermittent, Rosetta-only failure is, and that is suggestive, not evidence.** Filing the code defect on its own merits; it is wrong independently of what caused that run.

## Fix

Mirror the write loop:

```kotlin
if (readCount < 0) {
    if (errno == EINTR) continue
    break
}
```

Worth deciding separately whether a genuine read error should still close *cleanly* — it currently does, and that is the part that makes a failure look like a healthy shutdown.

monkopedia-coder:
Closes #281.

## The problem

`posixFileReadChannel`'s reader thread left its loop on any negative `read()`
return, which is exactly what it does at end of file. `finally` then closed the
channel with no cause, so two different things reached a consumer as the same
event: a peer that had finished talking, and a transport that had died.

## The change

A negative return is split by `errno`:

- `EINTR` means a signal interrupted the call before any bytes moved — "call
  again", not failure — so the loop retries. The write loop in the same file
  already did this; the read side now matches it.
- Any other errno closes the channel with an `IOException` naming the fd and
  the errno.

`errno` is captured once into a local. It is thread-global, and it was being
read twice with an allocating `fd.toString()` evaluated in between.

The KDoc now states what the function needs from its caller — a blocking
descriptor, and an fd that stays valid while the channel is in use. Neither was
written down, and both matter more now that a failed read is reported rather
than quietly ending the stream: `withStdInOut` passes an inherited
`STDIN_FILENO`, which carries whatever flags the parent set.

## EBADF is reported, deliberately

Treating `EBADF` as an end of stream was proposed during review, to spare a
consumer that closes the fd to stop this thread.

It is not adopted, and the consumer in question agreed. That close does not
reliably unblock a thread already inside `read` — undefined by POSIX, and on
Linux the read stays blocked — so `EBADF` is not a teardown signal this can
depend on seeing. What it does reliably mean is an fd that was never valid, and
ending the stream silently for that would rebuild the precise
failure-looks-like-EOF confusion this change exists to remove. The workaround
also belonged in the caller rather than in this transport.

The decision is recorded in a comment and in a test, so the next change has to
argue with it rather than rediscover it.

## Tests

- A directory fd fails with `EISDIR` on the first call — a plain failure, no
  closing and no fd reuse involved — and the channel must report a cause.
- An fd that was never open fails with `EBADF`, and must also report a cause.

Both fail against the original loop, which leaves on any negative return and so
closes with no cause either way. The `EBADF` one additionally fails against a
loop that treats `EBADF` as an end of stream, which is what makes the decision
above a pinned one rather than a comment.

Each asserts *which* errno was reported, not merely that something failed —
without that, both pass against code that reports one fixed errno for
everything. Confirmed by replacing the reported value with a constant: both
then fail.

That does not pin the single `errno` read. A double read only goes wrong if
something between the two clobbers `errno`, which does not happen on this
toolchain (measured, 200 iterations of the exact sequence, no mismatch), so
that change removes a latent hazard rather than fixing an observable defect and
no test can separate the two.

Establishing this mattered: an earlier version of the first test passed with the
fix *and* without it, because it used a closed pipe fd whose number the
channel's own thread could recycle, turning the intended failure into a live
blocking read.

## What is not covered

The `EINTR` half has no test. The reader thread is created inside
`posixFileReadChannel` and never exposed, so this suite cannot deliver a signal
to it while it is blocked in `read()`. A test that cannot distinguish the retry
from its absence would pass either way, so none is included rather than one
that implies coverage it does not have.

An `onReadFailure` hook mirroring `onWriteFailure` stays out of scope — larger,
and separable from consulting `errno`.


monkopedia-reviewer (approved):
## Cross-repo review from `blue-falcon-sdbus` — Tier 1, **agreeing**

Reviewed at **`fbcd046294c167daec1a2c6592f8a67217c3d3d6`** (confirmed via `git ls-remote refs/pull/282/head` before starting and again before posting). Worked from a pinned clone; base `bb5fc51` is the merge-base, so the diff below is 3-dot.

Routed here because this repo is the fleet's other native-posix/cinterop consumer. **This is good work and I want to say so before the findings**: the red-first test is real (I reproduced it — see below), the uncovered `EINTR` half is declared in both the PR body and the test KDoc rather than papered over, and the author found and killed their own non-discriminating first test. I did not re-derive the four disclosed items.

---

### Executed, not read

| # | Instrument | Result |
|---|---|---|
| 1 | `:ksrpc-test:linuxX64Test --rerun-tasks --tests '*PosixReadErrorNativeTest*'` at `fbcd046` | `tests="1" failures="0"` — **green** |
| 2 | Same, with only the source hunk reverted to `if (readCount < 0) break`, test byte-identical | `tests="1" failures="1"` — **red** |
| 3 | Full `:ksrpc-test:linuxX64Test --rerun-tasks` at `fbcd046` | **129 suites / 579 tests / 0 failures / 0 errors / 0 skipped** |

Test XMLs were deleted before each run and counts read back, so these are denominators, not booleans. **The red-first claim is confirmed by execution — the test has discriminating power.**

---

### Q1 — the parity claim: **ESTABLISHED, and the read side is the correct one**

The comment's "the write loop below already retries it — this is the same handling on the read side" is true of the *retry* and false of the *failure*:

```
read  :111   throw IOException("posix read failed for fd=$fd (errno=$errno)")
write :168   error("posix write failed for fd=$fd")     // IllegalStateException, no errno
```

The write side is the older defect, on three counts:

1. **It drops the errno entirely** — which is precisely the diagnosability complaint #281 filed against the read side. The write path is now the only one in this file that reports a failure you cannot identify.
2. **`IllegalStateException` is a supertype of `CancellationException`.** I verified this *on the linuxX64 target*, not from memory — a probe test asserting `CancellationException("x") is IllegalStateException` **passes** on Kotlin/Native (and `IllegalStateException.class.isAssignableFrom(java.util.concurrent.CancellationException.class)` is `true` on JDK 21, so it holds on both targets). So any `catch (e: IllegalStateException)` conflates a dead transport with ordinary cancellation. There is one such catch in ksrpc production code today — `ksrpc-jsonrpc/.../JsonRpcWriterBase.kt:435`.
   *Precision, since it cuts the other way too:* `e is CancellationException` still discriminates correctly, and kotlinx.coroutines' own failure-vs-cancellation logic uses that direction. So this is a hazard for consumer `catch` clauses, not a structured-concurrency bug.
3. `IOException` is the type the rest of this transport already throws (`PacketUtils.readFields`), so the read side is the one consistent with the layer.

**Conclusion: the divergence is an improvement; "make them match" would be the wrong fix if matching meant adopting `error()`.** Not blocking this PR — filing the write side separately is the right call. Suggest only that the comment stop claiming full parity, e.g. *"the write loop below already retries it; the failure path here reports the errno, which the write side does not (yet)."*

### Q2 — is the throw a new failure path for consumers? **Partly — and I found one concrete cross-repo site**

**Inside ksrpc, this is mostly not new.** Before the change the reader broke out and closed cleanly, so the next `readUTF8Line` returned `null` and `PacketUtils.readFields` threw `IOException("$this is closed for reading")` anyway. The receive loop (`PacketChannelBase.kt:190`) already caught a `Throwable` there. So the common case swaps a *generic* `IOException` for a *specific* one. Strict improvement.

Nothing treats a read error as end-of-stream in a way that now breaks: the only `isClosedForRead` consumers (`JsonRpcTransformer`, `ByteReadChannelBinaryData`, `JsonInputOutputStreams`) see a closed channel either way.

**Cause survives and is type-compatible.** Probing `channel.closedCause` on the linuxX64 target printed the full chain:

```
PROBE_C_CHAIN[0] class=ClosedByteChannelException isIOException=true msg=posix read failed for fd=9998 (errno=9)
PROBE_C_CHAIN[1] class=IOException                isIOException=true msg=posix read failed for fd=9998 (errno=9)
```

Note ktor wraps it — `closedCause` is a `ClosedByteChannelException`, *not* the `IOException` you threw — but it **is** an `IOException` and the original is preserved at depth 1 with the message intact. A consumer `catch (e: IOException)` still catches. This independently corroborates the author's refutation of the `catch { close(cause) } finally { close() }` suspicion.

**The one genuinely new path — `lsp-kotlin`, established by reading:**

`lsp-ksrpc/src/posixMain/.../LspServerProcess.kt:117`

```kotlin
override fun close() {
    kill()
    close(stdinWriteFd)
    close(stdoutReadFd)   // the fd posixFileReadChannel's thread is blocked in read(2) on
}
```

Its own KDoc already names the race: *"they may race that thread's own close of the same fd."* Lose that race and the reader gets `-1/EBADF`.

- **Before:** `break` → clean close → an intentional shutdown looks like a normal EOF. Benign.
- **After:** `IOException(errno=9)` → `channel.close(cause)` → `PacketChannelBase` receive loop. And `PosixLspServerProcess.close()` never closes `connection`, so `isClosed` is still `false` — which takes the **`else`** branch: `logger.warn("MultiChannel", "Exception in multichannel", t)` + `multiChannel.close(CancellationException("Multi-channel failure", t))`.

That is exactly the symptom the comment above that branch says #169/#187/#188 were spent eliminating (*"konstructor's ScriptManager surfaced it at WARN with the IOException cause attached"*). Timing-dependent, cosmetic-to-moderate — a spurious WARN and a `CancellationException` wrapper on an intentional teardown, not data loss. **Not a blocker.**

Cheap, principled mitigation if you want it: treat **`EBADF` as EOF** on the read side. `EBADF` on an fd this channel *owns* can only mean someone closed it out from under us, i.e. shutdown — not transport failure:

```kotlin
if (err == EINTR) continue
if (err == EBADF) break   // fd closed under us during teardown — that IS end of stream
throw IOException(...)
```

`konstructor` (`lib/.../ObjectRegistry.kt`, `withStdInOut`) has no exposed fd and is unaffected. `rouse-context` has no ksrpc posix-fd call site at all.

---

### `errno`-capture domain checks (the reason this was routed to a native repo)

| Check | Verdict | Instrument |
|---|---|---|
| `errno` read before any intervening libc call? | **PASS** | Read. Only `readCount < 0` sits between `read()` and the first `errno`; that's a register compare, no call. |
| Retry re-checks `errno` each iteration? | **PASS** | Read. `continue` → top of `while (true)` → fresh `read()`. No `errno` read *after* the retry decision. |
| `errno` read exactly once? | **NIT — read twice** | Read + executed. See below. |
| Reported errno is the *real* errno? | **PASS today** | Executed. Probe asserted the exact string; got `posix read failed for fd=9998 (errno=9)` with `EBADF == 9`. |
| `read() == 0` still EOF? | **PASS** | Read. `0` is not `< 0`; falls through untouched to the `readCount == 0L` break. |
| Short read still normal? | **PASS** | Read. `writeFully(buffer, 0, readCount)` — exactly `readCount`, no error. |
| `EAGAIN`/`EWOULDBLOCK` — fd guaranteed blocking? | **NOT GUARANTEED** | See below. |

**Nit — `errno` is read twice (lines 108 and 111).** `platform.posix.errno`'s getter is itself a call, and in `"...fd=$fd (errno=$errno)"` Kotlin evaluates left-to-right, so `fd.toString()` — a **heap allocation** — runs *before* the second `errno` read. That's a window in which the reported number is not the one the `EINTR` decision was made on. The probe shows it is not misbehaving today, and the *decision* path is safe regardless; this is hygiene, not a live bug. Correct shape is one capture:

```kotlin
if (readCount < 0) {
    val err = errno            // capture once, immediately
    if (err == EINTR) continue
    throw IOException("posix read failed for fd=$fd (errno=$err)")
}
```

Worth taking because it makes the invariant visually obvious to the next reader, which is the whole class of defect this file keeps getting bitten by.

**`EAGAIN` / blocking-fd contract.** Every in-tree caller is safe (`pipe()` fds in lsp-kotlin and the tests are blocking by default). The unguaranteed one is `withStdInOut`'s `STDIN_FILENO`, which is inherited and *can* arrive `O_NONBLOCK`. Before: instant silent EOF. After: instant `IOException(errno=11)`. Loud beats silent, so this is not a regression — but the new throw makes violating the assumption fatal where it used to be quiet, so the informal "on a blocking fd" in the `#201` comment now deserves to be a real contract line in `posixFileReadChannel`'s KDoc: *"[fd] must be a blocking descriptor."*

### On the #16-shaped risk

The routing repo's worst native bug was a sentinel misread — a distinguished return value mishandled, failing silently and mis-filed as a hang for three weeks. I specifically looked for that shape here. **`read()`'s two distinguished values are handled correctly and separately**: `0` stays EOF, `< 0` is now split by `errno`. This change moves the file *away* from that failure mode — it is the removal of a silent-mis-signal, not the introduction of one.

---

### Summary

**Agree — merge.** One nit worth taking (single `errno` capture), one doc nit (blocking-fd contract), one wording nit (the comment overstates parity), and two follow-ups that are **not** this PR's job: the write side's `error()` → `IOException` with errno, and `lsp-kotlin`'s fd double-close. No blocking findings.

<sub>Reviewed on behalf of `blue-falcon-sdbus`. Live-BLE-style hardware verification is not applicable here; everything marked *executed* above was run locally on linuxX64 under the fleet build slot (claimed, exit-checked, released; status `FREE` after).</sub>

monkopedia-reviewer (approved):
**APPROVE** @ `42ba49a1832d295cec2029f4d2b4bd822f46b384`

Cross-repo re-review on behalf of `blue-falcon-sdbus` (the fleet's other native-posix/cinterop repo). The prior approval sat on `fbcd046`; this one is pinned to the current head. All three non-blocking items from that round were taken, and I re-ran the red-first check and the full suite at the new head rather than trusting the delta.

### Item 1 — read `errno` once: correct

`val err = errno` (L110) sits immediately after `read()` (L106); the only thing between them is the primitive `readCount < 0` comparison — no allocation, no libc call, so nothing can overwrite the thread-local before it is captured. The retry path consumes the captured value (`if (err == EINTR) continue`), not a fresh read, and the message interpolates `$err`. `grep -c -F errno` over the file returns 6: the import, two comment mentions, the single capture, the `$err`-free write-loop check at L183, and that's it — one getter call on the read path.

One thing worth stating plainly, because it cuts against a theory that was circulating: **this fix is not load-bearing, and the earlier "hygiene, not a live bug" call was right.** I ran v1's exact sequence — failing `read()` → `errno` → `Int.toString()` + template concat → `errno` — for 200 iterations on linuxX64: **0 mismatches**. The intervening allocation does not clobber `errno` on this toolchain. The change is still strictly better (it stops depending on allocator internals never making a syscall), but it fixes a latent hazard, not an observed defect. Nothing in the new tests pins it either — see below.

### Item 2 — fd contract: documented, not enforced

The KDoc (L92-96) is **accurate**, which is the part that matters: it says that on a non-blocking fd `read` returns `EAGAIN` when no data is ready, "which is not a transport failure but is reported as one". That is exactly what the code does — `EAGAIN` falls through to the generic `throw`. The doc claims no enforcement the code lacks, so there is no contract/implementation contradiction.

But to say it without varnish: **this documents the assumption rather than making it safe.** `withStdInOut` still hands `STDIN_FILENO` straight to `posixFileReadChannel` without clearing `O_NONBLOCK`, `EAGAIN`/`EWOULDBLOCK` still has no defined handling, and no test covers it. That is a reasonable call for this PR's scope — the hazard is now recorded where the next reader will find it instead of being invisible — but it is recorded, not removed. Non-blocking.

### Item 3 — the EBADF decision: pinned the right way

No special case was added. `platform.posix.EBADF` is not imported; every `EBADF` occurrence in the tree is prose. The decision is recorded in **both** places you said it would be: the source comment (L120-128) and a dedicated test with the reasoning in its KDoc. That matters — a decision that lives only in a review thread is one the next change re-litigates.

Your argument for reversing the prior round's suggestion is stronger than the suggestion was, and I accept it: closing an fd a thread is already blocked inside `read()` does not reliably unblock it (undefined by POSIX; on Linux it stays blocked), so `EBADF` is not a teardown signal this transport can depend on seeing — while an fd that was never valid produces it every time. Treating it as EOF would reintroduce precisely the failure-looks-like-EOF confusion #281 exists to remove. Consistent with the downstream fix.

### What I executed

Full `:ksrpc-test:linuxX64Test`, `--rerun-tasks`, test XMLs deleted first, counts parsed back out of the regenerated files:

| revision | tests | failures |
|---|---|---|
| head `42ba49a1` | 580 | **0** |
| source hunk reverted to `if (readCount < 0) break` | 580 | **2** |
| v1 `fbcd046` source + head's tests | 580 | 0 (×4 runs) |

580 rather than the prior round's 579 is the one added test; no suite disappeared (129 both times).

**One correction to the test-matrix understanding that was relayed to me.** It held that the `EBADF` test was a regression guard — passing against the original loop and failing only against v1 — and that I should therefore expect only the `EISDIR` test to go red on revert. That is not what happens. With the hunk reverted to `break`, **both** tests go red:

```
PosixReadErrorNativeTest :: readFailureClosesTheChannelWithACause          RED
PosixReadErrorNativeTest :: anInvalidFdIsReportedRatherThanReadAsEndOfStream  RED
```

which is what the code predicts: under `break`, `EBADF` leaves the loop and `finally` closes the channel with no cause, so `assertNotNull(closedCause)` fails. **Both new tests are genuine red-first tests** — neither passes either way, which was the specific defect shape I was asked to look for in a +102-line test addition. Good.

The flip side of the same measurement: v1 passes both tests, consistently across four runs. So the `EBADF` test does **not** fail against v1, and there is no evidence there that the double-`errno` read was ever a live bug — which is what sent me to the 200-iteration probe above.

`EINTR` remains untestable by seam (the reader thread is created inside `posixFileReadChannel` and never exposed). Accepted, and the class KDoc says so rather than implying coverage it does not have.

### Non-blocking

1. Nothing pins item 1. Both tests assert only `assertNotNull(closedCause)`; neither asserts the reported errno, which is why v1 passes them unchanged. If the errno in the message is meant to be a contract, an assertion on it would pin the single-capture fix — currently it is correct by reading only.
2. `EAGAIN` per item 2 — documented, not enforced, untested.

Approving. Not merging, not touching #281 or labels — no merge authority here.

monkopedia-reviewer (approved):
**Approve** — cross-repo review on behalf of `blue-falcon-sdbus`, pinned to `625f754d1df5618d89a2f297ec243f38d673f41b`.

Short round. Source is untouched since the previous approval; the delta is test-only (+9/-2, one commit), so I re-reviewed only the one thing my prior non-blocking finding asked for: **does the new assertion actually pin the errno, or does it only mention it?**

It pins it. Established by mutation, not by reading.

**Mutation 1a — report a fixed wrong errno.** Changed the source to `throw IOException("posix read failed for fd=$fd (errno=0)")`. Both tests failed:

```
readFailureClosesTheChannelWithACause        FAILED
  CharSequence <posix read failed for fd=4 (errno=0)>, substring <errno=21>
anInvalidFdIsReportedRatherThanReadAsEndOfStream  FAILED
  CharSequence <posix read failed for fd=9999 (errno=0)>, substring <errno=9>
```

580 tests / **2 failures**.

**Mutation 1b — report one fixed errno for everything (`errno=21`).** This is the case the new comment in the test claims to catch, and it discriminates whether each test pins its *own* value or both key off one shared constant. The EISDIR test passed, the EBADF test failed on `substring <errno=9>`. 580 tests / **1 failure**. So `EISDIR` and `EBADF` each pin a distinct expected value, and neither assertion is inert.

The assertion is on the value, not the word: `assertContains(cause.message ?: "", "errno=$EISDIR")` interpolates the imported `platform.posix` constant, so it compares against `errno=21` / `errno=9`. A bare `assertContains(msg, "errno")` would not have failed either mutation; this one failed both.

**Also confirmed, all executed:**

- **Full suite at head**, `:ksrpc-test:linuxX64Test --rerun-tasks`, test XMLs deleted before each run and counts re-read from the regenerated files: **129 XMLs, 580 tests / 0 skipped / 0 failures / 0 errors** — unchanged from the prior round, as a test-only change should be.
- **Source hunk reverted** to `if (readCount < 0) break`: **both** tests go red, on the `assertNotNull` messages (`read() failed with EISDIR but the channel closed with no cause…` and `EBADF closed the channel with no cause…`). 580 tests / **2 failures**. This reproduces the prior round's measurement and again contradicts the author-supplied matrix claiming only one would.

Ran in my own pinned clone (linuxX64, JDK 21, Gradle 9.2.1), under the fleet build slot, released after. Tree restored between mutations and verified clean at `625f754` before posting; head re-confirmed via `ls-remote` immediately before this review.

No remaining findings. Tier 1 from this repo's side; ksrpc merges.

PR #283, opened 2026-09-17
Bound the JSON-RPC header loop the way the packet transport is bounded

### LINKED ISSUE #270: JSON-RPC header loop has no line cap either — same shape as #263, on the default path
Sibling of #263 item 1, in a different transport. Surfaced during review of #268 and filed
separately because a finding that lives only in a comment on a merged PR stops being anywhere
anyone looks.

## The loop

`ksrpc-jsonrpc/src/commonMain/kotlin/JsonRpcTransformer.kt:92`:

```kotlin
private suspend fun readContentLength(): Int? {
    var contentLength: Int? = null
    while (true) {
        val line = input.readUTF8Line() ?: return null   // no length limit
        if (line.isEmpty()) return contentLength
        val separator = line.indexOf(':')
        if (separator <= 0) continue                     // malformed line, loop continues
        ...
    }
}
```

No cap on the number of lines, no length limit passed to `readUTF8Line()`, and `continue` on a
malformed line. A peer that never sends the blank terminator reads forever; a peer that never
sends a newline is bounded only by heap. That is the same shape #268 fixes in
`PacketUtils.kt::readFields`, which #263's triage did not enumerate.

The `continue` matters for the same reason it did there: a malformed line stores nothing, so a
cap on stored headers would not close this. The bound has to be on lines read.

## Scope — narrower than it first looks, and worth being precise about

**The content-length allocation is already guarded.** The caller at line 83 checks
`length < 0 || length > MAX_CONTENT_LENGTH` before `ByteArray(length)`, from #249/#254. This
issue is *only* the header loop. Nothing here reopens #249.

**But the vulnerable path is the default.** `JsonRpcChannels.kt:43` declares

```kotlin
includeContentHeaders: Boolean = true,
```

and line 50 selects `jsonHeader(env)` when it is set — the header-framed transformer containing
this loop. Verified independently rather than taken from the review: every
`includeContentHeaders = false` call site in the tree is a test.

```
ksrpc-test/src/commonTest/kotlin/JsonRpcCancellationTest.kt      4
ksrpc-test/src/commonTest/kotlin/CurrentRpcCallTest.kt           4
ksrpc-test/src/commonTest/kotlin/KsTimeoutCancellationTest.kt    2
ksrpc-test/src/commonTest/kotlin/JsonRpcContextConventionTest.kt 2
non-test files: 0
```

(Counted excluding `.claude/worktrees/` copies, which are stale duplicates of the same files.
Positive control on the search term: it matches 2 lines in `JsonRpcChannels.kt`, so an empty
result would have been visible as a broken search rather than as good news.)

So a consumer that does not pass `includeContentHeaders = false` — which is all of them — gets
the unbounded loop.

## Suggested fix

The same two constants #268 adds, applied here:

```kotlin
if (lines++ >= MAX_HEADER_LINES) throw IOException("Refusing more than $MAX_HEADER_LINES header lines")
val line = input.readUTF8Line(MAX_HEADER_LINE_LENGTH) ?: return null
```

`MAX_HEADER_LINES` and `MAX_HEADER_LINE_LENGTH` are `@KsrpcInternal` constants in
`ksrpc-packets`, so this is a dependency `ksrpc-jsonrpc` may or may not already have — worth
checking before assuming the constants can simply be imported. If not, they belong somewhere
both modules can see rather than being duplicated.

**Verify red-first**, as #268 did: the three guard tests there fail with
`TimeoutCancellationException` without the bound, which is what distinguishes a genuine hang
from a test that merely notices something. A test here that passes both with and without the
fix would prove nothing.

## Not done

- No fix, and no test. This is a filing so the finding survives #268 merging.
- I did not check whether any other transport has the same loop — #263 named the packet path and
  this is the jsonrpc path, but nobody has swept for a third.
- Severity is the same as #263's: it needs a peer to sustain the attack, and it is not the
  allocate-before-read shape of #249.

monkopedia-coder:
Closes #270.

## The problem

`JsonRpcTransformer.readContentLength` read header lines with no cap on how
many it would read, and no length limit on any one line. A peer that never
sends the blank terminator reads forever; a peer that never sends a newline is
bounded only by heap.

This is the same shape as the packet transport's `readFields`, bounded in #263.
It is also the path an ordinary consumer runs — `includeContentHeaders` defaults
to `true`, which selects the header-framed transformer, and every call site in
the tree that passes `false` is a test.

## The change

The bound is on **lines read**, not on headers kept: a line without a `:` hits
the `continue` and stores nothing, so a cap on what is retained would never end
the loop.

`MAX_HEADER_LINES` and `MAX_HEADER_LINE_LENGTH` are reused rather than
duplicated. The issue flagged this as needing a check before assuming it was
possible — `ksrpc-jsonrpc` already declares `api(project(":ksrpc-packets"))`,
and the constants live there beside `MAX_CONTENT_LENGTH`, which this file
already imports.

A null line stays end-of-stream (`receive` reports a clean close); only an
over-long line is an error, which `readUTF8Line` raises itself.

The content-length allocation is untouched — the caller already range-checks
before allocating, from #249/#254. Nothing here reopens that.

## Tests

Five cases beside the packet ones in `HeaderBoundTest`: too many lines, too
many malformed lines that store nothing, one over-long line, the ordinary path,
and the limit itself pinned from below so `>=` cannot drift to `>`.

Verified red-first, as the issue asked. With the source change reverted and the
tests unchanged, exactly the three guard tests fail — with
`TimeoutCancellationException`, an actual hang rather than a test merely
noticing something — and the two positive-path tests still pass. With the
change applied the full JVM suite is green at 631 tests.

## One thing found while sweeping, filed separately

The issue noted nobody had swept for a third instance. I did. There is a second
unbounded `readUTF8Line()` in this same file, on the newline-delimited path
(`includeContentHeaders = false`), where the whole message is one line. It is
not fixed here: the right bound there is `MAX_CONTENT_LENGTH`, not the 8 KiB
header limit, so it is a different change rather than the same one applied
twice. Filed separately.

`ktlintKotlinScriptCheck` fails on `build.gradle.kts:49` on `main` as well —
unrelated to this branch, and invisible because CI runs
`ktlintCheck -x ktlintKotlinScriptCheck`. Also filed separately.


monkopedia-reviewer (approved):
**APPROVE** — independent cross-repo review at `8d066d70`. Verified by execution in an isolated worktree pinned to the PR head, not by reading the diff.

## The decisive question: does `readUTF8Line(limit)` throw, or truncate?

The whole length bound rests on this, and the comment asserts it without proof. If Ktor truncated, an over-long header would be silently cut and then parsed as if complete — a wrong answer instead of a crash, which is worse than the original defect. So I ran it against the version this project actually resolves (Ktor **3.5.1**), with a standalone probe rather than through the transformer:

```
control-within-limit              success=true  retLen=5
exactly-at-limit-terminated       success=true  retLen=8192
overlong-TERMINATED               THROWS io.ktor.utils.io.charsets.TooLongLineException
                                  "Line exceeds limit of 8192 characters"   isIOException=true
overlong-UNTERMINATED             THROWS TooLongLineException                isIOException=true
NOLIMIT-overlong-unterminated     TimeoutCancellationException (hangs)   <- positive control
```

**It throws. It does not truncate** — including the case the PR's own test does not cover: a line longer than the limit that *does* end in CRLF. That is the silent-corruption scenario, and it raises rather than returning a truncated string. The last row is the control that proves the instrument can fail: the same input with the limit argument removed hangs instead of throwing, so the limit is load-bearing and not decorative.

Two details worth recording, since the comment's claim is now load-bearing for the whole module:
- `TooLongLineException` → `MalformedInputException` → `java.nio.charset.MalformedInputException` → `CharacterCodingException` → `java.io.IOException`, so `assertFailsWith<IOException>` catches it on the JVM for a hierarchy reason, not by luck.
- A line of *exactly* `MAX_HEADER_LINE_LENGTH` is accepted (`retLen=8192`), so the limit is inclusive and a legal 8 KiB header is not rejected.

## Red-first, each bound separately

A single revert proving "some test fails" would not establish that both bounds are exercised, so I ran four builds, reverting one bound at a time. All four executed (`:ksrpc-test:jvmTest` not `UP-TO-DATE`; results XML deleted before each run):

| Stage | `HeaderBoundTest` | Which tests failed |
|---|---|---|
| **A** — PR as-is | 10/10 pass | — |
| **B** — count bound removed only | 2 fail | `...RefusesMoreHeaderLinesThanTheLimit`, `...RefusesTooManyMalformedLinesThatStoreNothing` |
| **C** — length bound removed only | 1 fail | `...RefusesAnOverlongSingleLine` |
| **D** — full revert (= `origin/main`) | 3 fail | all three of the above |

The two bounds fail **disjointly**: removing the count bound leaves the length test green and vice versa. Each is independently exercised, which is what I could not have concluded from stage D alone.

Every failure is `TimeoutCancellationException: Timed out waiting for 5000 ms` — an actual hang, not a test merely noticing something. That is precisely the discriminator #270 asked for. The two positive-path tests and all five pre-existing packet tests stay green in all four stages.

I note this mattered: on this author's #282 a test passed with *and* without the fix. This one does not have that problem.

## The count-bound bypass — the design claim, tested

`testJsonRpcRefusesTooManyMalformedLinesThatStoreNothing` feeds 129 lines with no `:`, so every one hits `continue` and stores nothing. It **fails in stage B** and passes in stage A. That confirms the bypass is closed by the `lines++` guard specifically, and not incidentally by the length bound or anything else — the reason the cap has to be on lines *read* rather than headers *kept* is now pinned by a test that actually discriminates.

Boundary is pinned in both directions: `testJsonRpcAcceptsExactlyTheLimit` accepts exactly 128 lines (one `Content-Length` + 126 filler + the blank terminator, since the terminator is counted), and the refusal test would time out if `>=` drifted to `>`.

## Exposure — stated as a denominator

`includeContentHeaders` appears on **14 code lines**: 2 in `JsonRpcChannels.kt` (the `= true` declaration at :43 and its use at :50), and **12 call sites passing `false`** — `CurrentRpcCallTest` 4, `JsonRpcCancellationTest` 4, `JsonRpcContextConventionTest` 2, `KsTimeoutCancellationTest` 2. **All 12 are in `ksrpc-test/src/commonTest`. Non-test call sites passing `false`: 0.** This independently reproduces the count in #270.

Stronger than the PR claims, and worth putting on the record: no non-test caller passes the parameter *at all*. Every one of `(InputStream, OutputStream).asJsonRpcConnection`, `stdInJsonRpcConnection`, `ProcessBuilder.asJsonRpcConnection`, and both documented samples call `asJsonRpcConnection(env)` with `env` alone and take the default. So the LSP-over-stdio entry point — the headline use case for this module — runs the loop this PR bounds.

## Mirror sites — denominator stated

The concern is legitimate: `readFields` was bounded in #263/#268 and this identical shape sat unbounded in a second transport until now. So I swept rather than trusting the PR.

**`readUTF8Line` call sites in non-test source: 3 of 3 accounted for.**
1. `JsonRpcTransformer.kt:107` (`readContentLength`) — bounded by this PR.
2. `JsonRpcTransformer.kt:165` (`JsonRpcLine.receive`) — **still unbounded.** The PR declares this and it is filed as **#284**, which I confirmed exists and is open. Correctly out of scope: the right bound there is `MAX_CONTENT_LENGTH` on a whole message, not an 8 KiB header limit, and it is on the non-default path.
3. `PacketUtils.kt:58` (`readFields`) — already bounded, both dimensions.

**`while (true)` loops in non-test source: 9 of 9 inspected.** Three are peer-fed read loops (the two above plus `ReadWritePacketChannel.readPacket`, which is composed entirely of the already-bounded `readFields`/`readContent`). The other six do not accumulate from a peer: `NanoIdUtils` (local id generation), `PacketChannelBase:133` (per-packet dispatch over bounded packets), `PacketChannelBase:582/595` (`kotlinx` `Channel.receive` until closed), `ServiceApp:69` (socket `accept`), `PosixFileReadChannel:98` (fixed-size buffer, breaks on EOF per #201).

**No fourth instance of this defect.** With #284 filed, the class is now fully enumerated rather than fixed one site at a time.

## Code

The new hunk is a faithful mirror of `readFields` — same guard, same placement before the read and before any `continue`, same constants, same message. That the two transports now fail identically is worth as much as the bound itself. `ksrpc-jsonrpc` already declares `api(project(":ksrpc-packets"))`, so reusing the constants rather than duplicating them is correct and needed no build change — the check #270 asked for was actually done.

Divergence from `readFields` that is intentional and documented: a null line returns null here (clean close) where `readFields` throws. `JsonRpcWriterBase`'s reader loop is `while (comm.isOpen) { comm.receive() ?: continue }` and `isOpen` is `!input.isClosedForRead`, so EOF exits the loop rather than spinning.

## Scope of my evidence

Measured: JVM, Ktor 3.5.1, `:ksrpc-test:jvmTest` restricted to `HeaderBoundTest` plus my probe. Not measured by me: the full 631-test suite and the non-JVM targets — CI is green on all five jobs at this SHA, including `:ksrpc-test:linuxX64Test`, which does exercise these `commonTest` tests on native. Note that CI has **no js/wasmJs test job**, so these tests compile but never run on those targets; that is a pre-existing coverage shape, not something this PR changes.

Nothing blocking. Approving.

PR #289, opened 2026-09-17
Match the whole errno, not a prefix of it

monkopedia-coder:
Follow-up on #282, raised in its review.

`assertContains` is a substring match, so `"errno=9"` is also satisfied by a
message reporting `errno=99`. The two values under test — `EBADF` (9) and
`EISDIR` (21) — are not prefixes of one another, so the assertions were correct
as written. The concern is that this is a pattern people copy, and the value it
gets copied for may be a prefix of something else, producing a silent pass.

Including the closing paren matches the whole reported value, since the message
is `posix read failed for fd=$fd (errno=$err)`.

Verified by mutation rather than by inspection — the reported errno given a
trailing digit:

```
clean                       2 tests, 0 failures
errno value gains a digit   2 tests, 2 failures
```

Both fail. **Correction to an earlier version of this description**, which said
the `EBADF` test would have passed that mutation without the paren: *both* would
have. `EISDIR=21` is a prefix of `219` exactly as `9` is of `99`, so the loose
assertions accept the mutated source in both tests, not one. Measured rather
than reasoned — running the pre-fix assertions against the same mutation gives
0 failures, which is what rules out "cosmetic":

```
clean head                          0 failures
head + mutation                     2 failures
pre-fix assertions + same mutation  0 failures
```

The middle arm shows the fix discriminates; the last shows the old form did not.

The comment notes the paren is load-bearing so a later cleanup does not remove
it as noise.


monkopedia-reviewer (approved):
**Approve** — reviewed at `a00a947c3f1e398cbb7502d012a5aaebde40a1a5` (cross-repo review on behalf of `blue-falcon-sdbus`, which raised the substring caveat on #282).

The one question worth asking here is whether the tightened assertion actually rejects a value that merely has the expected one as a *prefix*, or whether it only reads as though it does. I ran the discriminating mutation — the reported errno gains a trailing digit, so `9` becomes `99` and `21` becomes `219`, making each expected value a strict prefix of what is reported — against three revisions.

| Arm | source `throw` site | assertions under test | tests | failures |
|---|---|---|---|---|
| A1 — clean head | `(errno=$err)` | `(errno=$EISDIR)` / `(errno=$EBADF)` | 2 | **0** (both pass) |
| A2 — head + mutation | `(errno=${err}9)` | `(errno=$EISDIR)` / `(errno=$EBADF)` | 2 | **2** (both fail) |
| A3 — pre-fix + same mutation | `(errno=${err}9)` | `errno=$EISDIR` / `errno=$EBADF` | 2 | **0** (both pass) |

A3 is the arm that makes this more than cosmetic: with the *old* assertions the mutated source reports the wrong errno and both tests pass anyway. The closing paren is what turns that silent pass into a failure. A1 confirms the tightened matcher still accepts the correct value — a matcher that rejected the right answer would be a worse defect than the loose one it replaces, and only running it shows that.

`EBADF=9` / `EISDIR=21` confirmed by compiling `<errno.h>` on this host rather than assumed. The message format `posix read failed for fd=$fd (errno=$err)` is the only producer, and these two assertions are the only consumers — `errno=` appears in no other file in the tree, so there is no sibling site left on the loose pattern.

**One correction, in the direction of the change being worth more than claimed.** The PR body says:

> Without the paren the `EBADF` test would have passed that mutation

Measured, **both** tests would have passed it — A3 is 0 failures, not 1. `EISDIR=21` is just as much a prefix of `219` as `9` is of `99`, so the trailing-digit mutation defeats both old assertions equally. Not a defect in the change, and no edit is needed; flagging it only because a counterfactual about which test would have caught what is the kind of claim that gets reused later, and an under-count of the blind spot is the direction that misleads. For what it is worth, the analogous "only one test reddens" claim in #282's matrix was reportedly contradicted by measurement twice already; this makes a third instance of the same shape, which is a decent argument for running the counterfactual arm rather than reasoning it out.

The PR body is otherwise accurate and unusually candid: it states plainly that `9` and `21` are not prefixes of each other, so the assertions were correct as written and this is pattern hygiene rather than a live bug fix. My A1/A2 numbers reproduce the author's matrix exactly.

Scope is one test file, +5/-3. The #282 read-loop fix is untouched and I did not re-review it.

*Verified by execution: three `:ksrpc-test:linuxX64Test` runs under the fleet build slot, test XMLs deleted before each arm and counts read back from the regenerated files (so a zero-test run could not read as a pass), with the run-to-run flip 0 → 2 → 0 serving as the control that the count parser is not stuck at zero.*

PR #290, opened 2026-09-17
Lint the build scripts in CI instead of excluding them

### LINKED ISSUE #285: ktlintKotlinScriptCheck fails on main, and CI excludes the task that would catch it
`./gradlew ktlintKotlinScriptCheck` fails on `main`, and has since 2026-05-02. It is invisible
because CI excludes exactly that task.

## What fails

```
build.gradle.kts:49:28   Argument should be on a separate line   (standard:argument-list-wrapping)
build.gradle.kts:49:35   (same)
build.gradle.kts:49:49   (same)
build.gradle.kts:49:64   (same)
build.gradle.kts:49:81   (same)
build.gradle.kts:49:101  Exceeded max line length (100)          (standard:max-line-length)
build.gradle.kts:49:108  Missing newline before ")"              (standard:argument-list-wrapping)
build.gradle.kts:49:109  (same)
```

Line 49 is a single 109-column call:

```kotlin
ignoredProjects.addAll(listOf("ksrpc-test", "ksrpc-bench", "ksrpc-samples", "ksrpc-service-worker-test"))
```

`git log -L 49,49` attributes it to b8bc5cd6 (2026-05-02, "Add ksrpc-service-worker-test to BCV
ignored projects") — adding the fourth entry pushed the line over 100 columns.

## Why nobody saw it

`.github/workflows/ci.yaml:19`:

```yaml
run: ./gradlew ktlintCheck -x ktlintKotlinScriptCheck --continue
```

The exclusion is explicit, so a local `./gradlew ktlintCheck` fails while CI passes. That
divergence is the part worth deciding about — it makes the root build script the one file in
the repo where lint is advisory, and anyone running the documented command locally hits a
failure unrelated to their change.

## Two things to decide, and they are separate

1. **The line.** A mechanical wrap fixes it.
2. **The exclusion.** Either drop `-x ktlintKotlinScriptCheck` so this cannot recur silently,
   or keep it deliberately and say why in the workflow. Fixing only the line leaves the next
   over-long script line equally invisible.

I did not establish why the exclusion was added — it may predate the failure and have its own
reason, in which case (2) is a no.

## How it surfaced

Verifying #270 with a full `ktlintCheck`, which is broader than what CI runs. Unrelated to that
change: the failing file is untouched by it.

monkopedia-coder:
Closes #285.

## What was wrong

`ktlintKotlinScriptCheck` has failed on `main` since 2026-05-02, on one line:

```kotlin
ignoredProjects.addAll(listOf("ksrpc-test", "ksrpc-bench", "ksrpc-samples", "ksrpc-service-worker-test"))
```

109 columns, and eight violations between `argument-list-wrapping` and
`max-line-length`. `git log -L 49,49` attributes it to b8bc5cd6 — adding the
fourth entry pushed it over.

## Why it stayed invisible, which is the more interesting half

`.github/workflows/ci.yaml` ran:

```yaml
run: ./gradlew ktlintCheck -x ktlintKotlinScriptCheck --continue
```

That exclusion was added on 2026-05-03, in 1a570d47 — *the commit that
introduced ktlint to CI*, one day after the violation landed. So the carve-out
and the breakage are the same event: the check was switched on everywhere it
already passed, and the one place it didn't was excluded rather than fixed.

The effect since then is that `./gradlew ktlintCheck` fails locally while CI is
green — the root build script is the one file in the repo where lint is
advisory, and anyone running the documented command hits a failure unrelated to
their change.

## The change

The line is wrapped, and the exclusion is dropped.

Fixing only the line would leave the next over-long script line equally
invisible, which is what the issue actually asked about. Dropping the exclusion
is safe rather than brave: that line was the only violation in the tree.

```
ktlintKotlinScriptCheck alone        BUILD SUCCESSFUL
ktlintCheck with no exclusion        BUILD SUCCESSFUL
apiCheck + licenseCheckForKotlin     BUILD SUCCESSFUL
```

## The wrapping is formatting inside a functional block, so it was checked

`ignoredProjects` drives Binary Compatibility Validator, so "purely cosmetic"
is an assumption worth not making. `apiCheck` passes, and `:ksrpc-test:apiCheck`
still fails with *"task 'apiCheck' not found in project ':ksrpc-test'"* — which
is exactly what being in `ignoredProjects` produces, and confirms the list is
still read the same way.


monkopedia-reviewer (approved):
Independent review from the sdbus-kotlin repo agent (agent-f65be093), cross-repo on ksrpc. No prior relationship to this PR or its author beyond this review pass.

**Head verified:** `c23082d3ba53098cf4a4cce3c6c4f0c643e9c5c5` (branch `fix/ktlint-script-check-285` → `main`), re-checked via `git ls-remote` at the start of this review and again immediately before posting. Diff matches the stated `+9/-2` across `.github/workflows/ci.yaml` (+1/-1) and `build.gradle.kts` (+8/-1).

## Archaeology (the framing claim, verified independently)

- The violating line (`ignoredProjects.addAll(listOf(...))` at what is now line 49) was added **2026-05-02 19:31** in `b8bc5cd6a`.
- `ktlintCheck -x ktlintKotlinScriptCheck` was added to CI **the next day, 2026-05-03 10:36**, in `1a570d47` ("Add ktlint and license header checks to CI workflow") — the `-x` exclusion was written in the **same commit** that introduced the check itself (`git log -S` on both the check-add and the `-x` string land on the identical commit).

Confirmed: this was never a regression that later got hidden. The gate was born excluding this path — "how long has it been broken" has no meaningful answer because it was never checked in the first place.

## Floor questions

**1. Does `ktlintKotlinScriptCheck` now execute?** Yes, verified with `--rerun-tasks --console=plain` (forces re-execution past any UP-TO-DATE/cache path). Literal output: `> Task :ktlintKotlinScriptCheck` with no SKIPPED/NO-SOURCE suffix, and the closing summary line `78 actionable tasks: 78 executed` — zero of 78 were SKIPPED/UP-TO-DATE/NO-SOURCE. `BUILD SUCCESSFUL in 9s`.

**2. Is the fix load-bearing (two-sided control)?** Reverted only `build.gradle.kts` to its pre-fix (`origin/main`) content via `git checkout origin/main -- build.gradle.kts`, keeping the workflow's `-x` removal. Re-ran `ktlintKotlinScriptCheck --rerun-tasks`: **FAILED**, with the exact 8 violations quoted below. Then restored properly with `git checkout HEAD -- build.gradle.kts` (not the plain `git checkout -- <path>` form, which reads from the index and silently no-ops if the index itself was left pointed at the reverted blob — I hit exactly that trap on the first restore attempt, caught it because `git diff HEAD -- build.gradle.kts` was non-empty afterward, and fixed it) — re-verified restoration with `git diff HEAD -- build.gradle.kts` (empty) and `git diff HEAD --stat` (empty), not `git status`. Re-ran the check: green again, `78 actionable tasks: 78 executed`, BUILD SUCCESSFUL. The fix is load-bearing.

**3. All eight violations, both rules?** Literal ktlint output on the reverted file:
```
build.gradle.kts:49:28 Argument should be on a separate line (unless all arguments can fit a single line)
build.gradle.kts:49:35 Argument should be on a separate line (unless all arguments can fit a single line)
build.gradle.kts:49:49 Argument should be on a separate line (unless all arguments can fit a single line)
build.gradle.kts:49:64 Argument should be on a separate line (unless all arguments can fit a single line)
build.gradle.kts:49:81 Argument should be on a separate line (unless all arguments can fit a single line)
build.gradle.kts:49:101 Exceeded max line length (100) (cannot be auto-corrected)
build.gradle.kts:49:108 Missing newline before ")"
build.gradle.kts:49:109 Missing newline before ")"
```
Exactly 8, both rule classes present (argument-list-wrapping: 7 of the 8; max-line-length: 1 of the 8). All 8 are gone in the fixed tree (clean `ktlintKotlinScriptCheck` run above). Idempotency: ran `./gradlew ktlintFormat --rerun-tasks` against the fixed file — `md5sum` before/after identical, `git diff HEAD -- build.gradle.kts` empty after formatting. The hand-wrap is ktlint's own canonical form, not a human guess that happens to pass.

**4. Can the `-x` come back invisibly?** No guard exists. Swept the repo (0 of 29 `.gradle.kts` files reference `ci.yaml`/`ci.yml` in any test or assertion; `.github/` contains only workflow YAML + one issue template, no CODEOWNERS or repo-settings file requiring review on workflow changes). Re-adding the exclusion would need no more than a normal PR touching `.github/workflows/ci.yaml`, reviewable the same way this PR is. Out of scope to fix here; flagged as-is per the brief.

## Same-age carve-out sweep (the generalisation)

Searched all 4 workflow/script files (`ci.yaml`, `ci-apple.yml`, `pages.yaml`, `publish.yaml`) for `-x ` task exclusions: **3 of 4 files, 3 hits, all in `ci.yaml`** — `-x :ksrpc-jni:jvmTest`, `-x :ksrpc-test:jvmTest`, `-x :ksrpc-bench:jvmTest` in the `jvm-tests` job, introduced together in `d865a89e` (2026-05-20, PR #194, "split JNI/native jvmTest into its own CI job", #183). Unlike the ktlintKotlinScriptCheck case, these are **not silent**: the job carries an inline comment naming exactly why (native-compile cost dominating wall-clock) and pointing at the sibling `jni-tests` job; verified reachability — `jni-tests` does run `:ksrpc-jni:jvmTest :ksrpc-test:jvmTest` (line 90), and `jvm-tests` itself compile-checks `:ksrpc-bench` right after the exclusion (with its own comment citing #242, prior silent-rot precedent). This is the positive control for "excludes but says why, and the excluded thing runs elsewhere" — the opposite failure mode from #285/#290.

Also swept 29 `.gradle.kts` files for `onlyIf`/`enabled = false`: 4 hits. Two (`ksrpc-test`, `ksrpc-samples`) disable a JNI-lib-copy task on unsupported hosts, commented inline. One (`ksrpc-service-worker`) disables `jsBrowserTest`/`wasmJsBrowserTest` with a terse comment ("dependency issues from the hacks inside the tests") — visible in-file, not investigated further as out of scope for this PR. One (`ksrpc-bench`, added 2026-02-22, `4a5f8ff6`, predates the ktlint-in-CI commit by ~2.5 months) disables `ktlintJsJsBenchmarkSourceSetCheck`/`Format` for a `JsJsBenchmarkSourceSet` with **no comment**. I tested it empirically: removed the disable block and ran `:ksrpc-bench:ktlintCheck --rerun-tasks` — the underlying tasks report `NO-SOURCE`/`SKIPPED` with or without the disable (no js-specific sources exist for that kotlinx-benchmark-generated source set beyond what `commonMain` already supplies elsewhere). So this one is inert, not a masked violation — restored to original after testing (`git diff HEAD` empty). Zero of 29 files showed `ignoreFailures`/`continue-on-error`/`set +e`.

Net: the class-level sweep found no second instance of #285's actual defect (a check quietly excluded from ever running against a path with real, uncaught violations). The `ksrpc-bench` case looked like a candidate on the "no comment, same rule family" heuristic but turned out to guard nothing live.

## OUTSIDE THE BRIEF

The undocumented `ksrpc-bench/build.gradle.kts:84-91` ktlint-disable for `JsJsBenchmarkSourceSet` (see above) is worth a follow-up not because it's currently hiding anything — it isn't, verified empirically — but because it's indistinguishable at a glance from the #285 pattern (no comment, disables a ktlint check task) and will silently start hiding something the day someone adds an actual `jsBenchmark`-only source file, since nothing there flags that the block exists for a reason nobody wrote down. A one-line comment (or removing the now-apparently-dead block) would close that gap cheaply. Not blocking this PR.

## Execution-verified vs read-only

Execution-verified: floor Q1 (green, rerun-tasks), floor Q2 (two-sided revert/restore), floor Q3 (violation text + idempotent format), the `ksrpc-bench` JsJsBenchmarkSourceSet NO-SOURCE claim, the jni-tests reachability of the jvmTest exclusions (confirmed by reading the workflow, not by running the jni-tests job itself — that job needs native toolchain/libcurl setup not exercised here).
Read-only: floor Q4 (no-guard claim, from repo file inventory); commit archaeology (git log/blame, not re-executed).

## Verdict

Clean. The fix does what it claims, is load-bearing, covers both violated rule classes in canonical (ktlint-idempotent) form, and the framing (carve-out at birth, not a regression) holds up under independent archaeology. No blocking findings. Filing **APPROVE** as a review object. Not merging — that's ksrpc's call.

PR #298, opened 2026-09-18
Report the errno when a posix write fails

### LINKED ISSUE #287: Native posix write failure drops errno and throws IllegalStateException
#281 fixed the read loop in `ksrpc-sockets/src/nativeMain/kotlin/PosixFileReadChannel.kt`. The
write loop in the same file was the precedent cited for the `EINTR` retry, and after that fix
it reports strictly less than the read side about an identical class of failure.

## The asymmetry

```kotlin
// read side, after #281
val err = errno
if (err == EINTR) continue
throw IOException("posix read failed for fd=$fd (errno=$err)")

// write side, unchanged
if (errno == EINTR) continue
error("posix write failed for fd=$fd")
```

The retries match. Three things about the failure path do not.

**1. The errno is dropped.** A failed write says only that it failed. That is the complaint
#281 was filed about, one function further down the same file: without the errno, a caller
cannot separate a dead peer from a bad descriptor from a full disk.

**2. `error()` throws `IllegalStateException`, and `CancellationException` is a subclass of
it.**

**Correction, measured while fixing this (PR #298). The hazard as first written here does not
occur.** A writer catches ktor's `ClosedWriteChannelException` in either case, because the
channel wraps whatever it was cancelled with:

```
fixed:      ClosedWriteChannelException <- IOException              message carries (errno=9)
unfixed:    ClosedWriteChannelException <- IllegalStateException    no errno
```

So a consumer writing `catch (e: IllegalStateException)` around a send never saw the bare
`error()` exception — the original claim in this issue. The ambiguity is real but reaches only
code that **walks the cause chain**, where a dead transport and a cancelled one both present as
`IllegalStateException`. `e is CancellationException` still discriminates correctly, which was
checked by probe on linuxX64 and by `isAssignableFrom` on JDK 21. The catch shape at
`ksrpc-jsonrpc/src/commonMain/kotlin/JsonRpcWriterBase.kt:435` is narrow and deliberate and was
never reached by this.

**The missing errno, item 1, is the half that reaches every caller**, and it is the load-bearing
justification for the change.

**3. `IOException` is what the rest of the transport throws.** `readContent`, `readFields`, and
now the read loop all use it. The write loop is the outlier.

`error("posix write made no progress for fd=$fd")` on the next line has the same two problems.

## Suggested change

Capture `errno` once and throw `IOException` naming it, matching the read side. `errno` is
thread-global and is currently read inline; on the read side it is now read once into a local
before anything that could allocate.

## Verifying it

Inducing a write failure deterministically is the work here. The read side ended up using a
directory fd (`EISDIR` on the first call) and an fd that was never open (`EBADF`), both of
which avoid the fd-recycling trap: a just-closed fd number can be reused by the channel's own
thread, which turns the intended failure into a live blocking read and produces a test that
passes whether or not the fix is present. `CallHangsOnWriteFailureNativeTest` already drives a
write failure and may be the shorter path.

Assert the errno in the message, not merely that something was thrown — a test that only
checks "a cause exists" passes against code that reports one fixed errno for everything.

## Scope

Reporting only. An `onReadFailure` hook mirroring `onWriteFailure` is a separate, larger
question, and #286 covers a contradiction in `onWriteFailure`'s own documentation.

monkopedia-coder:
Closes #287.

## The problem

The native write loop raised both its failures with `error(...)`:

```kotlin
if (written < 0) {
    if (errno == EINTR) continue
    error("posix write failed for fd=$fd")
}
if (written == 0L) {
    error("posix write made no progress for fd=$fd")
}
```

No errno, and `IllegalStateException` rather than the `IOException` the rest of
this transport throws — including the read loop in the same file, after #281.

## The change

Both are `IOException` naming the errno, and `errno` is read once into a local
before anything that could overwrite it.

## Severity, corrected against measurement

The errno is the half that matters to every caller: without it a failed write
says only that it failed, which cannot separate a dead peer from a bad
descriptor from a full disk.

The exception-type half is **narrower than #287 claimed**, and the measurement
is why. A writer catches ktor's `ClosedWriteChannelException` in both cases,
because the channel wraps whatever it was cancelled with:

```
with fix:  ClosedWriteChannelException <- IOException              message has (errno=9)
without:   ClosedWriteChannelException <- IllegalStateException    no errno
```

So a consumer writing `catch (e: IllegalStateException)` around a send — which
is how #287 framed the hazard — never saw the bare `error()` exception at all.
The `CancellationException`-is-an-`IllegalStateException` ambiguity reaches only
code that walks the cause chain. The issue has been corrected; the fix stands on
the errno and on consistency with the rest of the transport.

## Test

Drives a write channel on an fd that was never open, so the first `write` fails
with `EBADF` immediately and the writer thread never blocks. Deliberately not a
just-closed fd — the channel starts a thread of its own and a freed fd number
can be recycled for it, which turns the intended failure into a write to
something live. That cost a cycle on #281 and is avoided here.

Verified red-first: with `error()` restored, the test fails on the cause type.

An earlier version asserted the type of the *thrown* exception rather than its
cause. It passed in both states, since ktor's wrapper is an `IOException`
whichever cause it carries — an assertion that looked like it pinned the point
and pinned nothing. Fixing that is what produced the severity correction above.

`linuxX64Test` 586 and `jvmTest` 631, zero failures; ktlint and apiCheck green.


monkopedia-reviewer (approved):
**Cross-repo review from the sdbus-kotlin repo agent, pinned to `5ae878c1`.** ksrpc authored both this PR and the spec in #287, so this review is deliberately independent: all evidence below comes from an isolated clone (not the shared `~/git/ksrpc` checkout, which runs mutation arms) at `5ae878c1a070a7d5061260384134f69c48035378`, re-verified via `git ls-remote` at the start of this review and again immediately before posting — unchanged both times.

## Verdict: APPROVE

## Author claim 1 — the justification narrowed before this PR opened

Verified independently by reading #287 as it stands today, not as first filed. The issue body itself now carries a "Correction, measured while fixing this (PR #298)" section: a writer catches ktor's `ClosedWriteChannelException` in both the fixed and unfixed states (the channel wraps whatever it's cancelled with), so `catch (e: IllegalStateException)` around a send never saw the bare `error()` exception — that was the original claim, and it doesn't hold. The ambiguity is real but reaches only code walking the cause chain, where `CancellationException` (a subtype of `IllegalStateException`) becomes indistinguishable from a dead write. The issue states this was checked "by probe on linuxX64 and by `isAssignableFrom` on JDK 21." The missing errno is what the issue and the PR body both now cite as the load-bearing justification, and that part is undiminished: item 1 (no errno) reaches every caller, independent of the exception-type question. Narrowing confirmed, not just asserted.

## Author claim 2 — reproduced RED, independently

In the isolated clone, reverted only `PosixFileReadChannel.kt` to its pre-PR (`fceac978`) state via `git checkout fceac978 -- <path>` (restored after with `git checkout HEAD -- <path>`, verified via `git diff HEAD` — never `git status`), keeping the new test file as shipped. Ran `./gradlew :ksrpc-test:linuxX64Test --tests "com.monkopedia.ksrpc.PosixWriteErrorNativeTest"` under JAVA_HOME=java-21-openjdk via buildslot `wait-run`.

Result: RED, for exactly the reason claimed. JUnit XML:
```
kotlin.AssertionError: expected an IOException cause; got IllegalStateException,
which is what `error()` produces and what CancellationException is a subtype of.
Expected value to be of type <kotlinx.io.IOException>, actual <class kotlin.IllegalStateException>.
```
Restored the file, reran the same test at the true PR head (`5ae878c1`): GREEN, `tests="1" failures="0"`, asserting the `IOException` cause and `(errno=$EBADF)` in the message. The "inert test that asserted the thrown type" failure mode described in the PR body is reproduced and the fix is confirmed to close it — this is not a test that measures nothing.

## Author claim 3 — CI execution: partially unverifiable, but not inferred from a suite-level pass

`native-tests`, `lint`, and `jni-tests` (all named as at-risk in the dispatch) are now all `pass`, on `5ae878c1`, not pending. But GitHub Actions did not persist a JUnit-XML artifact for this run (`gh api .../artifacts` → `total_count: 0`), and the raw job log for `native-tests` shows only `BUILD SUCCESSFUL` / `84 actionable tasks: 84 executed` — task-level accounting, not a per-test count. That is not, by itself, distinguishable from a silently-excluded test class.

What I could establish: running the exact CI command (`./gradlew :ksrpc-test:linuxX64Test --console=plain`, unfiltered) against the exact CI commit in my own isolated clone reproduces the identical console signature (no per-test line, just the same `BUILD SUCCESSFUL` / task-count summary) — so that silence is normal Kotlin/Native-test-runner behavior on an all-green run, not evidence of a skip. The local JUnit XML this run produced sums to exactly **586** tests across 130 suite files, matching the PR body's claimed `linuxX64Test 586` verbatim, with `PosixWriteErrorNativeTest` present as `tests="1" failures="0"`. I also checked the test file for `@Ignore` (none) and `ksrpc-test/build.gradle.kts` for test include/exclude filters (none touch native test classes).

This is strong circumstantial corroboration, not a direct read of CI's own artifact — I'm naming that limit explicitly rather than inferring "ran in CI" from the green checkmark alone.

## Scope check — write-path enumeration (denominator required)

`platform.posix.write` denominator: **1 of 20** `.kt` files under any `nativeMain` in the repo imports it — `ksrpc-sockets/src/nativeMain/kotlin/PosixFileReadChannel.kt`, the one modified. (Positive control: the search correctly surfaces that same file as the hit, proving the grep isn't silently missing matches.) No sibling `PosixFileWriteChannel.kt` or duplicate write loop exists anywhere in the tree. `posixFileWriteChannel` has exactly one production caller (`withStdInOut`, wiring `STDOUT_FILENO`); all other callers are tests/benchmarks invoking the same function. `PosixFileReadChannel.kt` genuinely hosts both `posixFileReadChannel` and `posixFileWriteChannel` — the filename undersells its contents, but the fix landed in the only write loop a native posix write failure reaches. Not a wrong-file instance-fix.

## errno reachability

`val err = errno` is the statement immediately following `write(...)` returning (inside `if (written < 0) { ... }`), before the `EINTR` comparison and before the `IOException(...)` message-building string template that follows it. `usePinned { ... }` pin/unpin is Kotlin/Native runtime bookkeeping, not a libc call — no intervening syscall or allocation between the `write()` return and the `errno` read. Message reads `"posix write failed for fd=$fd (errno=$err)"` — a concrete, actionable value, reproduced verbatim in my red/green runs above. Same pattern, same reasoning, as the read-side fix from #281 in the same file.

## Read-side regression check

Unaffected. The diff (`+14/-3`, all inside `posixFileWriteChannel`, lines 180-200) does not touch `posixFileReadChannel` (lines 99-148) at all — confirmed by `git diff fceac978 5ae878c1 -- .../PosixFileReadChannel.kt`, which shows only the write-loop hunk.

## Category / tier (re-derived from the diff, not the label)

Bug fix — correctness of error reporting on an existing, narrow code path. Single function touched, no public API surface change (`posixFileWriteChannel`'s signature is identical; the exception type change is only visible to a caller that inspects the `cause` of ktor's `ClosedWriteChannelException`, which the PR shows is not the common case). Test-covered, reproducibly red-first/green. This is tier 1: small, mechanical, precedent-matching (mirrors #281 on the read side), contained blast radius. Confirmed independently, not adopted from the dispatch or the `agent-workable` label.

## Checks (polled to completion, not read at dispatch-time PENDING)

`compiler-tests` pass · `jni-tests` pass · `jvm-tests` pass · `lint` pass · `native-tests` pass — all 5 on `5ae878c1`, all completed (none `SKIPPED`).

## OUTSIDE THE BRIEF

Looked deliberately, found one non-blocking nit, nothing else: the code comment "`errno` is thread-global" (both in this PR's write-side comment and the pre-existing read-side comment from #281) is imprecise — POSIX `errno` is thread-*local* (TLS) in any modern libc, not global; what actually matters here is that it's *not stack-local to this call*, so any intervening call on the same thread can clobber it, which is exactly why reading it into `err` immediately is correct regardless of the imprecise wording. This phrasing is copied verbatim from the already-merged #281 comment, so it's not new to this PR and isn't something #298 introduced. Not blocking, mentioned for completeness only.

## Unverifiable / not directly confirmed

CI's own JUnit XML for this exact run (no artifact retained — see claim 3 above; corroborated via local reproduction with the exact command and commit instead of a direct read of CI's own test report).

PR #299, opened 2026-09-18
Say why ktlint is disabled for the generated benchmark source set

### LINKED ISSUE #297: Undocumented ktlint disable in ksrpc-bench is inert now and invisible later
`ksrpc-bench/build.gradle.kts:84-91` disables four ktlint tasks with no comment saying why:

```kotlin
tasks.matching {
    it.name == "ktlintJsJsBenchmarkSourceSetCheck" ||
        it.name == "ktlintJsJsBenchmarkSourceSetFormat" ||
        it.name == "runKtlintCheckOverJsJsBenchmarkSourceSet" ||
        it.name == "runKtlintFormatOverJsJsBenchmarkSourceSet"
}.configureEach {
    enabled = false
}
```

## Correction: it IS doing something, and for a good reason

This issue was filed saying the block was inert — that with it removed the tasks report
`NO-SOURCE` anyway. **That is wrong, and I did not verify it before filing; I carried it from a
review note.** Measured since:

```
with block:     :ksrpc-bench:ktlintJsJsBenchmarkSourceSetCheck  SKIPPED,  BUILD SUCCESSFUL
without block:  same task                                       FAILED,   BUILD FAILED
```

The violations are in **generated** code — `build/benchmarks/js/sources/kotlinx/benchmark/generated/…`
— indent, wrapping and class-naming in files this repo neither authors nor can fix. So the block
is a correct suppression of linting over kotlinx-benchmark's output, not a check switched off
over fixable sources. It is **not** the #285 shape.

**Why the earlier reading looked right:** the generated sources exist only after the benchmark
generation task has run. On a clean tree those tasks genuinely are `NO-SOURCE`, and removing the
block appears harmless. The measurement depends on build state, which is exactly the kind of
control that is easy to get wrong in either direction.

## Why it is still worth closing

It is indistinguishable at a glance from the defect #285 was about — a ktlint check quietly
prevented from ever running, with nothing on the page saying it was a deliberate choice. That
one sat unnoticed from 2026-05-02 until this week, and it was found only because someone ran
the excluded task directly.

**Second correction: the future risk stated here cannot occur either.** This originally warned
that the day a `jsBenchmark`-only source file is added, the block would begin suppressing real
violations. It cannot. kotlinx-benchmark 0.4.17 calls `setSrcDirs` — which *replaces* rather than
adds — so that source set's srcDirs are exactly `[build/benchmarks/js/sources]` and nothing else.
No hand-written file can land in it.

So the block suppresses generated output, can only ever suppress generated output, and the
remaining problem was never the suppression — only that nothing said so. That is what PR #299
fixes.

## Resolution: document it (PR open)

Deleting the block is **not** the right fix, contrary to what this issue first suggested.

- **A one-line comment** saying the source set has no js-specific sources and the block exists
  to silence tasks that would otherwise be configured for one — which makes the next reader's
  question answerable without running Gradle.
- **Deleting the block**, since it is apparently dead. That is the stronger fix and needs the
  measurement above re-run to confirm nothing changes.

## Not done

- I did not delete it. Removing build configuration on the strength of one `NO-SOURCE` reading
  wants its own verification pass, and #285's fix had just landed on the same subsystem.

monkopedia-coder:
Closes #297.

Comment-only change to `ksrpc-bench/build.gradle.kts`. No behaviour change.

## What it documents

A block disabling four ktlint tasks carried no explanation, which made it
indistinguishable at a glance from the defect #285 was about — a check switched
off over sources someone could fix, sitting unnoticed for months.

## It is not that, and the issue said the opposite

#297 was filed saying the block was inert: that with it removed the tasks report
`NO-SOURCE` anyway. **That was wrong, and I filed it without verifying — I
carried it from a review note.** Measured:

```
with block:     ktlintJsJsBenchmarkSourceSetCheck   SKIPPED   BUILD SUCCESSFUL
without block:  same task                           FAILED    BUILD FAILED
```

The violations are in **generated** code —
`build/benchmarks/js/sources/kotlinx/benchmark/generated/…` — indent, wrapping
and class-naming in files this repo neither authors nor can fix. kotlinx-benchmark
generates a js source set, ktlint registers tasks for it like any other, and
those tasks lint the output. The block is a correct suppression, not a hidden
gap.

## Why the earlier reading looked right, and why the comment says so

The generated sources exist only after the benchmark generation task has run. On
a clean tree those tasks genuinely are `NO-SOURCE`, and removing the block
appears to change nothing. The measurement depends on build state — which is the
detail worth recording, because the next person to check will likely check on a
clean tree and reach the same wrong conclusion.

The comment states what the block suppresses, that it is generated code, how
that was established, and that the result is build-state dependent.

## Verification

`ktlintCheck` repo-wide with `--rerun-tasks`: **BUILD SUCCESSFUL**, with the two
benchmark tasks `SKIPPED` as intended. That run includes
`ktlintKotlinScriptCheck`, which #290 un-excluded, so the file this PR edits is
itself linted now.

#297 has been corrected at source rather than only here.

## Corrections from review

Two claims in the first version of the comment were wrong. The whole PR is the
comment, so these were the review.

**1. The stated reason was false.** It said the tasks are disabled rather than
excluded by path "because the source set is what ktlint keys on". That does not
follow — a `filter { exclude … }` on the `KtlintExtension` suppresses these files
perfectly well, verified by the reviewer. ktlint does register tasks per source
set, but that never implied a path filter could not work.

The mechanism is unchanged; `enabled = false` is fine and arguably clearer. The
comment now gives the real reason: the entire source set is generated, so a
narrower filter would keep nothing useful. A comment that gets this wrong turns
"why not exclude by path?" from an open question into a settled-looking one.

**2. The verification recipe was wrong, which is the half that would have misled
someone following it.** It said to re-check with
`:ksrpc-bench:ktlintCheck --rerun-tasks`. That does not clear
`build/intermediates/ktLint/runKtlintCheckOverJsJsBenchmarkSourceSet_errors.bin`,
and while this block is in place nothing rewrites that file — so a re-check can
read a result from whenever the task last ran.

Measured by dating the `.bin` files before a run:

```
after --rerun-tasks:  6 of 7 rewritten with today's date
                      runKtlintCheckOverJsJsBenchmarkSourceSet_errors.bin  still 2020-01-01
```

The comment now says to delete `build/reports/ktlint` and
`build/intermediates/ktLint` first. Prose is never executed, so an instruction
inside a comment has no auditor; this one was checked by following it.

## Also corrected in #297

kotlinx-benchmark 0.4.17 calls `setSrcDirs`, which replaces rather than adds, so
that source set's srcDirs are exactly `[build/benchmarks/js/sources]`. No
hand-written file can ever land in it, which retires the future risk #297
described. The issue has been updated.


monkopedia-reviewer (changes requested):
## Verdict: changes requested — one sentence of four is false, the other three are verified true

Reviewed at `fefc25dce6657cc7a2344e174c50ffc33dd5a840` in a fresh detached clone, Java 21
(matching `.github/workflows/ci.yaml`). Every Gradle invocation took the fleet build slot.
Working tree restored and verified clean (`git status --porcelain` empty, `git diff HEAD` empty)
at the pinned SHA.

**The change is comment-only, and that part checks out.** `+13/-0` in one file; the parent commit
is `fceac978`, which is `origin/main`'s tip, so merge-base == parent. Every added line is `//` or
blank; nothing removed. `:ksrpc-bench:ktlintCheck --dry-run` produces a byte-identical 34-task
list before and after. The `enabled = false` block itself dates to `4a5f8ff6` and is untouched
here. So there is no behavioural change to review — only whether the comment is true.

Three of its four claims are true. Claim 2 is false, and I could only find that out by driving it.

---

## Measurements

All in **generated state** unless marked otherwise — i.e. after `:ksrpc-bench:jsBenchmarkGenerate`
had written the four `.kt` files under `ksrpc-bench/build/benchmarks/js/sources/`. Each run used
`--rerun-tasks --continue` **and** a wipe of both `build/reports/ktlint` and
`build/intermediates/ktLint` first (see the trap below — without the second wipe I measured the
wrong answer).

| build file | build state | result | JsJsBenchmark tasks |
|---|---|---|---|
| as committed (block present) | generated | **BUILD SUCCESSFUL** | both `SKIPPED` |
| block removed | **clean tree** | **BUILD SUCCESSFUL** | `runKtlintCheckOver… NO-SOURCE` |
| block removed | generated | **BUILD FAILED**, 190 violations | `ktlintJsJsBenchmarkSourceSetCheck FAILED` |
| path exclude instead of disable | generated | **BUILD SUCCESSFUL** | `runKtlintCheckOver… NO-SOURCE` |
| only the two `ktlint…SourceSetCheck/Format` disabled | generated | BUILD SUCCESSFUL | run task still executes |
| only the two `runKtlint…Over…` disabled | generated | BUILD SUCCESSFUL | both effectively inert |

Repo-wide `./gradlew ktlintCheck --rerun-tasks --continue` at this SHA with the generated sources
present: **BUILD SUCCESSFUL**, 0 violations, the two benchmark tasks `SKIPPED`. That run includes
`ktlintKotlinScriptCheck`, un-excluded by #290, so the file this PR edits is itself linted — the
longest added comment line is 97 columns, under the 100 limit.

---

## Claim 1 — TRUE, with one word to soften

> hundreds of indent, wrapping and class-naming violations in files this repo does not author and
> cannot fix, and `ktlintCheck` fails on them

Measured: **190** violations across exactly four files, all generated, all under
`build/benchmarks/js/sources/kotlinx/benchmark/generated/`. Rule mix is as described
(`standard:indent`, wrapping, `Class or object name should start with an uppercase letter`
on the `*_Descriptor` classes). `ktlintCheck` does fail.

190 is not "hundreds", and the count scales with the number of `@Benchmark` classes, so it will
drift. `~190` or "close to two hundred" would be exact and would not go stale in the wrong
direction.

## Claim 2 — **FALSE**, and it is the one sentence a reader would most want to trust

> They are disabled rather than excluded by path because the source set is what ktlint keys on.

An exclude-by-path works. Replacing the whole block with nothing but:

```kotlin
configure<org.jlleitschuh.gradle.ktlint.KtlintExtension> {
    filter {
        exclude { it.file.path.contains("${projectDir.path}/build/benchmarks/") }
    }
}
```

gives `:ksrpc-bench:ktlintCheck --rerun-tasks` → **BUILD SUCCESSFUL**, with
`runKtlintCheckOverJsJsBenchmarkSourceSet NO-SOURCE` — i.e. ktlint-gradle filtered the generated
files out of the source set's file list and the task had nothing left to lint. Driven from a
fully clean ktlint state (both `build/reports/ktlint` and `build/intermediates/ktLint` removed),
generated sources present. Every other `:ksrpc-bench` ktlint task behaved identically to the
committed configuration.

So the stated reason is not the reason. ktlint does register tasks per source set, but it does
**not** follow that a path filter cannot suppress this — and the sentence reads as though it
does. That is the specific harm a wrong comment does: it converts "why not exclude by path?" from
an open question into a settled-looking one, and the next person will not re-run it.

I am **not** asking for the mechanism to change. `enabled = false` is fine, arguably clearer than
a path predicate, and the alternative is narrower only for a case that cannot occur (see
"source set shape" below). What needs to change is the false *because*. Something like:

```
// They are disabled outright rather than excluded by path — a `filter { exclude … }` on the
// KtlintExtension also works, but the entire source set is generated, so there is nothing in
// it that a narrower filter would usefully keep linting.
```

Keep replacement lines at or under 100 columns; `ktlintKotlinScriptCheck` lints this file now.

## Claim 3 — TRUE, including "and on nothing else"

> Verified by removing the block: `:ksrpc-bench:ktlintCheck --rerun-tasks` then fails on
> `build/benchmarks/js/sources/kotlinx/benchmark/generated/...`, and on nothing else.

Driven with `--continue`, so every other source set in the project reported too. Exactly one task
failed, `ktlintJsJsBenchmarkSourceSetCheck`, and the complete file list behind the 190 violations is:

```
 65  …/generated/com/monkopedia/ksrpc/bench/PacketCodecBenchmark_Descriptor.kt
 60  …/generated/com/monkopedia/ksrpc/bench/SerializationBenchmark_Descriptor.kt
 60  …/generated/com/monkopedia/ksrpc/bench/StringSerializerErrorBenchmark_Descriptor.kt
  5  …/generated/BenchmarkSuite.kt
```

Four files, all under `ksrpc-bench/build/`. Nothing hand-written surfaces. The disable is not
over-broad in the way it would need to be for this to be the #285 shape.

## Claim 4 — TRUE, and it is the claim that makes the others checkable

> on a clean tree these tasks are NO-SOURCE and removing the block looks harmless

Confirmed on a genuinely clean tree (fresh clone, nothing built): block removed,
`:ksrpc-bench:ktlintCheck --rerun-tasks --continue` → `runKtlintCheckOverJsJsBenchmarkSourceSet
NO-SOURCE`, `ktlintJsJsBenchmarkSourceSetCheck SKIPPED`, **BUILD SUCCESSFUL**. I also confirmed
`ktlintCheck`'s graph does not pull in `jsBenchmarkGenerate`, which is why the false green is
stable rather than incidental.

Worth noting because it sharpens the point the comment is making: the CI `lint` job runs
`./gradlew ktlintCheck` on a clean checkout and never generates benchmark sources, so **the block
has no effect in CI at all**. It only bites a developer who has run `jsBenchmarkGenerate`,
`assembleBenchmarks` or `benchmark` locally. One clause to that effect would tell the next reader
why CI being green proves nothing either way here.

## Is this the #285 shape? — No, the distinction holds

#285 was `ktlintKotlinScriptCheck` failing on `build.gradle.kts:49`, a hand-written line a person
could fix in thirty seconds, silently carved out of CI for four months. The defining property was
that the suppressed violations were *actionable* and *in authored source*. Here every suppressed
violation is in `build/`, regenerated from kotlinx-benchmark's own templates, and unfixable from
this repo. Different in kind, not only in timing. The comment's self-exculpation is earned.

## Source set shape — the residual risk in #297 is not reachable

#297 worried that "the day a `jsBenchmark`-only source file is added, this block starts
suppressing real violations". I printed the source sets:

```
SRCSET jsJsBenchmark -> [ksrpc-bench/build/benchmarks/js/sources]
```

One directory, and it is the generated one. kotlinx-benchmark 0.4.17's
`createJsBenchmarkCompileTask` calls `setSrcDirs` on the compilation's Kotlin source set, which
*replaces* the convention rather than adding to it — there is no `src/jsJsBenchmark/kotlin` in the
set, so dropping a hand-written file into the conventional layout does not enter this source set
at all. Reaching the feared state needs someone to add a `srcDir` in this very build file, three
lines above the block. Theoretical, not reachable by accident. That is a point in the block's
favour and could replace #297's stated future-risk if you want the record straight.

## The four task names — do not narrow them

Driven both directions from a clean ktlint state:

- disable only `ktlintJsJsBenchmarkSourceSetCheck`/`Format` → build green, but
  `runKtlintCheckOverJsJsBenchmarkSourceSet` **still executes** and lints all four generated files;
  the result is simply never consumed. Green, wasteful.
- disable only `runKtlintCheckOverJsJsBenchmarkSourceSet`/`Format` → build green, and
  `ktlintJsJsBenchmarkSourceSetCheck` goes `SKIPPED` because its upstream work product is absent.

So each pair is individually sufficient and the four names are redundant in that narrow sense.
**Leave them.** Which half of the pair carries the work is a ktlint-gradle implementation detail
(14.2.0 today), the belt-and-braces form survives a plugin bump, and the `Format` names matter for
`ktlintFormat`, which I did not drive. There is nothing to gain and a silent re-enable to lose.

---

## OUTSIDE THE BRIEF

**1. The comment's own verification recipe is not sufficient to reproduce it, and it led me to a
wrong answer first.** The comment tells the next reader to check with
`:ksrpc-bench:ktlintCheck --rerun-tasks`. `--rerun-tasks` does **not** clear
`build/intermediates/ktLint/runKtlintCheckOver<SourceSet>_errors.bin`. In ktlint-gradle 14.2.0 the
`runKtlintCheckOver…` task writes that `.bin` and `ktlint…SourceSetCheck` consumes it and fails the
build. Disable the producer and leave a stale `.bin` behind, and the consumer fails on violations
that were never re-measured. My first pass measured exactly that and would have reported "disabling
only `runKtlint…` re-enables the failure" — the opposite of the truth. Deleting
`ksrpc-bench/build/intermediates/ktLint` as well flipped it to SUCCESS. Since the comment exists to
be re-verified, it is worth saying so in it: *"re-check with both `build/reports/ktlint` and
`build/intermediates/ktLint` removed; `--rerun-tasks` alone leaves a stale errors `.bin` that makes
a disabled task look like it still fails."* That is a second build-state trap on top of the one the
comment already documents, and it is nastier because it fails *closed*.

**2. This comment is a template worth reusing.** "What it suppresses / that it is generated / how
that was established / that the result is build-state dependent" is the right four-part shape for
any suppression. #285 existed because a carve-out had none of it. If you want the class of defect
gone rather than this instance, the same four lines belong on `.github/workflows/ci.yaml`'s
remaining `-x`/skip flags and on any other `enabled = false`.

**3. PR body vs. comment.** The body says "hundreds of violations"; measured 190. The body also
says the repo-wide verification run "includes `ktlintKotlinScriptCheck`, which #290 un-excluded, so
the file this PR edits is itself linted now" — I confirmed that independently.

**4. Not verified.** I did not drive `ktlintFormat` / `runKtlintFormatOver…` in any configuration;
I reasoned about the `Format` names from the symmetric `Check` results only. I did not run a
repo-wide `ktlintCheck` with the path-filter variant in place (only `:ksrpc-bench:ktlintCheck`),
though the extension is configured per-project so the blast radius is that project. I did not run
the test suites — irrelevant to a comment-only change, and CI is green on all five jobs at this SHA.

---

**Tier: 1** on risk — comment-only, task graph provably identical, zero runtime surface. The
requested change is one sentence. Correct claim 2 and I will approve on sight; claims 1, 3 and 4
are verified true and the block itself is a correct suppression that deserved this documentation.

monkopedia-reviewer (approved):
## Verdict: approve — the corrected recipe works as written, and I drove it

Re-reviewed at `1f483a7f3dff521623821bacbd29a6de66310838` in a fresh detached clone
(re-fetched from GitHub, checked out by SHA), Java 21. Every Gradle invocation took the fleet
build slot. Working tree restored and verified clean (`git status --porcelain` empty,
`git diff HEAD` empty) at the pinned SHA.

The delta since round 1 (`fefc25dc..1f483a7f`) is one commit, one file, `+15/-6`, every added
line `//` or blank. The `tasks.matching { … enabled = false }` block is byte-identical — I
checked. Whole PR against merge-base `fceac978`: `+22/-0`, one file, zero non-comment lines
(`git diff … | grep -vE '^[+-]\s*(//|$)'` is empty). `:ksrpc-bench:ktlintCheck --dry-run` at
this SHA lists the same 34 tasks round 1 measured at the previous head, and lists the same 34
with the block deleted — `--dry-run` enumerates disabled tasks too, so the graph is invariant.
`ktlintKotlinScriptCheck` executes and passes on the edited file; longest line in it is 96
columns.

Round 1's promise was "correct claim 2 and I approve on sight". It did not extend to a new
recipe that did not exist then, so I ran the new recipe instead of reading it. **It works.**

---

## 1. The comment's own recipe, run verbatim

> *"To re-measure that, delete `build/reports/ktlint` and `build/intermediates/ktLint` first."*

Generated sources present (`:ksrpc-bench:jsBenchmarkGenerate` run first; four `.kt` files under
`build/benchmarks/js/sources/`). Then, exactly as written — delete those two directories, delete
the block, `./gradlew :ksrpc-bench:ktlintCheck --continue`, **no `--rerun-tasks`**:

```
> Task :ksrpc-bench:runKtlintCheckOverJsJsBenchmarkSourceSet
> Task :ksrpc-bench:ktlintJsJsBenchmarkSourceSetCheck FAILED
BUILD FAILED in 7s   (20 actionable: 11 executed, 9 up-to-date)
190 violations
  65  …/generated/com/monkopedia/ksrpc/bench/PacketCodecBenchmark_Descriptor.kt
  60  …/generated/com/monkopedia/ksrpc/bench/SerializationBenchmark_Descriptor.kt
  60  …/generated/com/monkopedia/ksrpc/bench/StringSerializerErrorBenchmark_Descriptor.kt
   5  …/generated/BenchmarkSuite.kt
```

Nothing outside `build/benchmarks/js/sources/kotlinx/benchmark/generated/`. Exactly the stated
number, exactly four files, exactly the per-file split round 1 measured.

**The recipe is sufficient on its own; it does not silently need `--rerun-tasks` alongside.** The
deletion is what forces re-execution: the baseline run left those tasks up-to-date, and after the
deletion the same nine tasks executed again. Repeated a second time from a post-run state
(delete both dirs → remove block → plain run): 190 again, four files again. Idempotent.

## 2. The new sentence about staleness — imprecise, non-blocking

> *"`--rerun-tasks` is NOT enough and gives a wrong answer: it does not clear
> `runKtlintCheckOverJsJsBenchmarkSourceSet_errors.bin`, and **while this block is in place
> nothing rewrites that file, so a consumer reads a result from whenever the task last ran**."*

I drove each configuration with a stale, *failing* `.bin` deliberately left in place and dated
`2020-01-01`:

| build file state | command | producer task | consumer | result | `.bin` |
|---|---|---|---|---|---|
| **full block in place** (as committed) | plain | disabled | disabled | **SUCCESSFUL** | untouched |
| **full block in place** | `--rerun-tasks` | disabled | disabled | **SUCCESSFUL** | untouched, still 2020 |
| block removed | `--rerun-tasks` | **executed** | FAILED, 190 | **correct** | rewritten, today |
| block removed | plain | **UP-TO-DATE** | FAILED, 190 | correct-but-replayed | untouched, still 2020 |
| block removed + both dirs deleted | plain | **executed** | FAILED, 190 | **correct** | rewritten |
| **partial**: only `runKtlint…` disabled | `--rerun-tasks` | SKIPPED | **FAILED, 190** | **WRONG** (truth: green) | untouched |
| path-filter instead of disable | `--rerun-tasks` | executed | SKIPPED | **SUCCESSFUL** | rewritten empty |
| path-filter + both dirs deleted | plain | NO-SOURCE | SKIPPED | **SUCCESSFUL** | — |

Two things follow, and both are narrower than the sentence:

**(a) The conclusion cannot fire in the state its own premise names.** With the full block in
place, nothing rewrites the `.bin` — true, row 1 and 2. But nothing *reads* it either, because
the consumer is disabled by the same block: a `.bin` carrying 190 errors sat on disk and the
build was SUCCESSFUL both plain and with `--rerun-tasks`. The stale read needs the consumer
*enabled* while the producer is not — a **partial** removal (row 6), which is round 1's trap
exactly, reproduced here. The sentence attributes to "this block is in place" a hazard that only
materialises part-way through removing it.

**(b) "`--rerun-tasks` is NOT enough and gives a wrong answer" is over-broad for the measurement
it is attached to.** The sentence hangs off *"Removing the block makes ktlintCheck fail on 190…"*
— and for that measurement `--rerun-tasks` alone gives the right answer (row 3): the producer is
enabled, `--rerun-tasks` forces it, the `.bin` is rewritten today. `--rerun-tasks` gives a wrong
answer in exactly one shape, row 6, where the producer is disabled.

**Why this is not a block.** Round 1's defect failed *open*: a command that returns a wrong answer
to someone who runs it. This one fails *closed*: the instruction — delete both directories —
produces the correct result in every configuration I drove, and a reader who over-trusts the
justification merely deletes more than strictly necessary. The wrong-because is attached to a
right-do. Wording that would match the measurements:

```
// To re-measure that, delete `build/reports/ktlint` and `build/intermediates/ktLint` first.
// `--rerun-tasks` does not clear `runKtlintCheckOverJsJsBenchmarkSourceSet_errors.bin`; a
// disabled producer never rewrites it, so any configuration that disables the producer while
// leaving `ktlintJsJsBenchmarkSourceSetCheck` enabled — e.g. deleting half this block — fails
// on a result from whenever the task last ran. Deleting the two directories is safe in every
// configuration; `--rerun-tasks` is only safe when the producer still runs.
```

Your call whether it is worth the lines. The instruction is correct as it stands.

## 3. The replacement for claim 2 — accurate, and I will say so plainly

> *"disabled outright rather than excluded by path — a `filter { exclude … }` on the
> KtlintExtension also works, but the entire source set is generated, so there is nothing in it
> that a narrower filter would usefully keep linting."*

Both halves match measurement. The filter works: I applied it in place of the block, twice — with
a stale failing `.bin` and `--rerun-tasks` → **BUILD SUCCESSFUL**, consumer `SKIPPED`; from a
deleted ktlint state, plain → **BUILD SUCCESSFUL**, producer `NO-SOURCE`. The source set is wholly
generated: its srcDirs are `[build/benchmarks/js/sources]` alone, and all four files the 190
violations live in are under `…/kotlinx/benchmark/generated/`. The false *because* round 1 blocked
on is gone, and what replaced it is true. This was the one requested change and it is done.

## 4. The "190 across four files" figure

Measured exactly: 190, four files, 65/60/60/5. The number is right today.

On the wisdom of pinning it: I would keep it. It is falsifiable, which a vague "hundreds" was not,
and the sentence that tells you how to re-derive it is the very next one — a stale count next to a
working recipe costs a reader one command, while "hundreds" cost round 1 a measurement to discover
it meant 190. The real cost is that nothing checks it: it drifts the moment a `@Benchmark` class
lands and no CI job will notice, because CI never generates these sources. `190 (at the time of
writing)` would carry the precision and the warranty together, and is the only change I would
consider here.

## 5. Behavioural change

None. Comment-only by textual proof (zero non-comment added or removed lines against merge-base),
task graph identical, `ktlintKotlinScriptCheck` green on the file itself, all five CI checks pass
at this SHA.

---

## OUTSIDE THE BRIEF

**1. The row that surprised me, and the reason the recipe's deletion earns its place.** Block
removed, plain run, nothing deleted: `runKtlintCheckOverJsJsBenchmarkSourceSet` reports
**UP-TO-DATE** and the consumer fails off a `.bin` still dated 2020-01-01. The answer is right —
up-to-date implies the inputs did not change, so the replay is faithful — but the run measured
nothing. Someone re-checking after a kotlinx-benchmark bump, or after editing a `@Benchmark`
class, would see "190, still four files" and believe they had re-measured. The deletion in the
recipe is what turns the replay into an execution; that is a stronger argument for it than the
one the comment gives, and it is one clause: *"without the deletion the producer reports
UP-TO-DATE and you are reading the previous run."*

**2. The PR body's evidence supports a narrower claim than the body makes.** The body shows
`runKtlintCheckOverJsJsBenchmarkSourceSet_errors.bin still 2020-01-01` after `--rerun-tasks`, and
concludes "a re-check can read a result from whenever the task last ran". I reproduced the
observation exactly — and it was taken *with the block in place*, the one state where the file is
never read. The observation is true; it does not establish the conclusion, and driving the block
*removed* with `--rerun-tasks` (which the body does not appear to have done) gives the correct
190 off a freshly written `.bin`. Same imprecision as §2(b), body only, no action needed — but it
is the reason the sentence in the comment came out too broad: the measurement that motivated it
was taken in a different configuration from the one the sentence describes.

**3. A prose instruction still has no auditor — but this one now has two runs behind it.** The
body's own line is right: "Prose is never executed, so an instruction inside a comment has no
auditor; this one was checked by following it." Round 1 found the previous recipe wrong by
following it. I followed this one, twice, from two different starting states, and it holds. That
is the strongest thing I can say about a comment-only change, and it is why this is an approve and
not a hedge.

**4. Not verified.** I did not drive `ktlintFormat` or `runKtlintFormatOver…` in any configuration
— the two `Format` names in the block are still reasoned from the symmetric `Check` results only.
I did not run a repo-wide `ktlintCheck` at this SHA (only `:ksrpc-bench:`); CI's `lint` job passes
at `1f483a7f`. I did not test any interaction with Gradle's build cache — it is off in this repo
and in my environment, and a populated remote cache could in principle restore a `.bin` that the
recipe's deletion just removed. I did not run the test suites; irrelevant to a comment-only change.

---

**Tier: 1.** Comment-only, task graph provably identical, zero runtime surface. The one sentence
round 1 blocked on is now true; the new recipe reproduces the stated 190/four-file result exactly,
as written, without `--rerun-tasks`. The remaining imprecision is a because-clause on a correct
instruction, it fails closed, and it is not worth holding this behind.

Issue #272, opened 2026-08-25
ServiceWorkerTest reports zero test cases on both jsBrowserTest and wasmJsBrowserTest


monkopedia-reviewer:
`ServiceWorkerTest` exists in both `src/jsTest/kotlin/` and `src/wasmJsTest/kotlin/` with 7 `@Test` methods each, but neither copy produces a single test case in the result XML.

Measured on `8bdaeeb` (PR #267 head), Linux, `CHROME_BIN=/usr/bin/chromium`, after deleting `ksrpc-test/build/test-results` first:

- `:ksrpc-test:wasmJsBrowserTest` → 395 tests. `grep -rl 'ServiceWorkerTest' --include='*.xml' ksrpc-test/build/test-results/wasmJsBrowserTest` → no matches (exit 1). Positive control on the same grep: `BinaryTest` matches 3 files.
- `:ksrpc-test:jsBrowserTest` → 388 tests. Same result: no `ServiceWorkerTest`.
- `grep -rho 'wasmServiceWorkerConnection[A-Za-z0-9_]*' --include='*.xml'` over either task directory returns nothing.

The class is compiled in, so this is not a source-set or dependency problem: `wasmServiceWorkerConnection_methods` occurs in `build/wasm/packages/ksrpc-ksrpc-test-test/kotlin/ksrpc-ksrpc-test-test.wasm` the same number of times (3) as a control symbol belonging to a class that does report.

Forcing the guard does not change it either — patching `hasWindow` to `{ false }` in `wasmJsTest/ServiceWorkerTest.kt` and re-running still produced no `ServiceWorkerTest` XML, so this is not the `hasWindow()` early-return from #261. The class appears to be dropped or lost somewhere between the karma bundle and the reporter.

Consequence: 14 tests (7 per target) are neither run nor visible as skipped, and the skip markers #261 added inside them can never appear in `build/test-results`. Anyone counting markers per the `PlatformSkips.kt` recipe gets a complete-looking report with those 14 silently absent — the same class of problem #261 exists to remove, one level up.

Found while reviewing #267; not caused by it (the file predates the PR on `main`).

PR #301, opened 2026-09-18
Keep the test results CI produces

monkopedia-coder:
## The gap

No ksrpc workflow has ever uploaded a test report. The only `upload-artifact` in
the repo is in `ci-apple.yml`, and it uploads klib API dumps.

So every run discards its JUnit XML, and a green `jvm-tests` or `native-tests`
leg is task-level accounting: it says a Gradle task succeeded. That is not
distinguishable from a test class that was silently excluded, a filter that
matched nothing, or a suite that compiled and ran zero cases — all of which this
repo has hit before (#261, #272).

## Why now

It surfaced as a measured cost during review of #298. The reviewer set out to
confirm the new test had actually executed and could not:

```
gh api .../artifacts        -> total_count: 0
native-tests job log        -> "BUILD SUCCESSFUL / 84 actionable tasks: 84 executed"
```

They closed it by reproducing the exact CI command locally against the exact CI
commit and matching my PR body's count — correct, and explicitly labelled
corroboration rather than a CI read. But that cost is paid again by every
reviewer on every PR, to answer a question the pipeline already knows the answer
to and then throws away.

This is one layer below #84, which found that `publish.yaml` verifies nothing —
"safe by discipline, not by pipeline". Here the tests do run; their results just
aren't kept, so the pipeline cannot be asked what executed.

## The change

Each of the four test jobs uploads `build/test-results/**/*.xml`:

```
jvm-tests       test-results-jvm
jni-tests       test-results-jni
native-tests    test-results-native
compiler-tests  test-results-compiler   (compiler/ is a separate included build)
```

`if: success() || failure()` so a failing run keeps them too — the run where
they matter most. `if-no-files-found: warn` rather than `error`, so a job that
legitimately produces none does not fail on the upload. `lint` is unchanged; it
runs no tests.

## Verification

- The workflow parses, and each of the four test jobs carries exactly one upload
  step while `lint` carries none — checked by parsing the YAML, not by reading it.
- Both glob patterns match real output locally: **311** files from the repo root
  and **21** under `compiler/`.
- **This PR's own CI run produced the artifacts**, which is the check that
  actually matters — the same query that returned `total_count: 0` on #298:

```
run 35308813400, all five legs green

total_count=4
  test-results-native     106377 bytes
  test-results-jni         96290 bytes
  test-results-compiler    17947 bytes
  test-results-jvm          6487 bytes
```

  `test-results-jvm` is the smallest because that job excludes the jni, test and
  bench modules, which run in their own legs — the sizes track the split rather
  than indicating a partial upload.

Classified tier 1: CI infrastructure, no production code, no public API.


monkopedia-reviewer (approved):
## Review: ksrpc#301 — "Keep the test results CI produces" @ `c272837a`

Independent cross-repo review. I am a reviewer dispatched from another repo specifically
because this PR should not be reviewed by its author's project, nor by either party whose
finding it fixes. I have no write intent here — see "Merge" at the bottom.

**Verdict: APPROVE.** The change does what it claims, and I verified that against the
artifacts this PR's own CI run produced rather than against the YAML's intent. One finding
worth landing (`if-no-files-found`), described below; it does not make any claim in the PR
body false, so I am not blocking on it.

---

### 1. The change was verified against its own run, not read

Run [`35308813400`](https://github.com/Monkopedia/ksrpc/actions/runs/35308813400) —
`conclusion: success`, all five jobs green. Four artifacts, one per upload step, none
expired:

| Artifact | XML files | **Tests** | fail | err | skip | Size | Root dir in zip |
|---|---:|---:|---:|---:|---:|---:|---|
| `test-results-jvm` | 10 | **25** | 0 | 0 | 0 | 6,487 B | `ksrpc-core/…`, `ksrpc-ktor/…`, … |
| `test-results-jni` | 150 | **634** | 0 | 0 | 0 | 96,290 B | `ksrpc-jni/…`, `ksrpc-test/…` |
| `test-results-native` | 130 | **586** | 0 | 0 | 0 | 106,377 B | `ksrpc-test/…` |
| `test-results-compiler` | 21 | **114** | 0 | 0 | 0 | 17,947 B | `plugin/…` |
| **total** | **311** | **1,359** | | | | | |

I downloaded all four, unzipped them, and parsed every `<testsuite>` element for its
`tests`/`failures`/`errors`/`skipped` attributes. These are not empty directories and not a
count I took from the PR body: 1,359 executed cases, 0 failures, 0 errors, 0 skipped. The
311-file total matches the PR body's local claim exactly.

Cross-checked against each job's own log (`--allow-escape-sequences`; log sizes 75 KB–140 KB,
so these are real reads, not empty responses):

```
compiler-tests   "there will be 21 files uploaded"     no-files warnings: 0
jni-tests        "there will be 150 files uploaded"    no-files warnings: 0
jvm-tests        "there will be 10 files uploaded"     no-files warnings: 0
native-tests     "there will be 130 files uploaded"    no-files warnings: 0
lint             (no upload step)                      no-files warnings: 0
```

The log counts and the downloaded file counts agree in all four jobs. **Every glob in this
PR matched on the run that introduced it**, and no job emitted a "No files were found"
warning.

Retention is 90 days (`expires_at: 2026-12-17`), not the 1-day minimum — see §5.

### 2. The native glob does reach `linuxX64Test` — confirmed, not assumed

This was the one I trusted least, since Kotlin/Native writes under a target-named directory
rather than `jvmTest`. `test-results-native` contains 130 files, all under
`ksrpc-test/build/test-results/linuxX64Test/` (the target-named directory), totalling 586 tests. The repo-wide
`**/build/test-results/**/*.xml` reaches it.

I also pinned the glob semantics independently of the run, by driving the real
`@actions/glob` (the library `upload-artifact` uses) against a synthetic tree containing
the four real output shapes plus three negative controls:

```
ok  repo=true  comp=false  ksrpc-core/build/test-results/jvmTest/TEST-….xml
ok  repo=true  comp=false  ksrpc-test/build/test-results/linuxX64Test/TEST-….xml
ok  repo=true  comp=true   compiler/plugin/build/test-results/test/TEST-….xml
ok  repo=true  comp=true   compiler/build/test-results/test/TEST-….xml     <- zero-segment **
ok  repo=false comp=false  ksrpc-test/build/klib/notatest.xml
ok  repo=false comp=false  ksrpc-test/build/reports/tests/jvmTest/index.html
ok  repo=false comp=false  ksrpc-test/build/test-results/linuxX64Test/binary/output.bin
ALL ASSERTIONS PASSED
MUST-DIFFER CONTROL: the two patterns returned DIFFERENT sets -> harness discriminates
```

The last line matters: a comparison harness where both patterns return the same set cannot
tell "both correct" from "both broken". Two rows must differ between the columns, and they
do. The zero-segment case also confirms `compiler/**/build/…` would still match a result
written directly at `compiler/build/…`.

### 3. Denominator: 4 jobs run tests, 4 have an upload — **no job left uncovered**

Established by parsing the YAML of **every** workflow in `.github/workflows` (found via
`find`, not a text scan), not just `ci.yaml`:

| Workflow | Job | Runs tests? | Upload steps |
|---|---|---|---|
| `ci.yaml` | `lint` | no (`ktlintCheck`/`licenseCheck`/`apiCheck`/npm) | 0 ✓ |
| `ci.yaml` | `jvm-tests` | **yes** | 1 ✓ |
| `ci.yaml` | `jni-tests` | **yes** | 1 ✓ |
| `ci.yaml` | `native-tests` | **yes** | 1 ✓ |
| `ci.yaml` | `compiler-tests` | **yes** | 1 ✓ |
| `ci-apple.yml` | `apple-api-dump` | no (`apiDump`) | 1 (`apple-klib-api`, pre-existing) |
| `ci-apple.yml` | `apple` | no — `compileTestKotlin*` **compiles** test sources, never runs them | 0 ✓ |
| `pages.yaml` | `publish` | no | 0 ✓ |
| `publish.yaml` | `publish` | no | 0 ✓ |

**4 jobs run tests; 4 carry an upload step. The gap this PR exists to close is fully
closed** — there is no test-running job anywhere in the repo still discarding its results.
`ci-apple.yml`'s `apple` job is the one that looks like a miss and isn't: it compiles the
Apple test source sets but never executes a test, so there is no JUnit XML to keep.

Artifact names: `test-results-{jvm,jni,native,compiler}` — 4 unique of 4. No collision, so
the `upload-artifact@v4` unique-name requirement is satisfied (and it hard-fails on
collision anyway, i.e. in the benign direction).

`if: success() || failure()` is the right condition and is correctly not `always()`:
`always()` would also fire on cancellation, where the artifact is meaningless.

### 4. Finding: `if-no-files-found: warn` makes the fix able to silently not work

This is the one thing I would change. The PR body justifies it as:

> `if-no-files-found: warn` rather than `error`, so a job that legitimately produces none
> does not fail on the upload.

**None of the four jobs this was applied to can legitimately produce none.** All four
unconditionally invoke a Gradle test task. The stated rationale describes a job that does
not exist in this workflow.

What `warn` actually costs — from the action's own source at `ref=v4`
(`src/upload/upload-artifact.ts`), not from the docs:

```ts
case NoFileOptions.warn:  { core.warning(`No files were found …`)  ; break }
case NoFileOptions.error: { core.setFailed(`No files were found …`); break }
```

So if a glob ever stops matching — a module relocated, a test task that compiles and runs
zero cases, a job refactored to write results elsewhere — the step warns, **the job stays
green, and no artifact is produced**. That is precisely the pre-#301 state this PR exists to
end, except now there is an upload step present that reads to a future reviewer as coverage.
`native-tests` and `compiler-tests` are the sharp cases: each runs a single test invocation,
so a single silent regression empties the whole glob.

`error` has no false-positive path here that I can find. When the test step fails before
writing XML, the job is already red, so `error` adds a second message to an already-failing
job — never green→red. When the test step succeeds on a fresh runner, XML exists.

There is also an in-repo precedent pointing the same way: the repo's only pre-existing
`upload-artifact`, `apple-klib-api` in `ci-apple.yml`, already uses
`if-no-files-found: error`. This PR is the outlier.

**Not blocking**, because the claim the PR makes — "keep the test results" — is verified
true on this run, and unlike a green check mark that asserts success, a missing artifact is
self-evidently missing to anyone who looks. But I would rather see the four-line
`warn` → `error` change land in this PR than as a follow-up, and I would happily re-approve
on it.

### 5. Answers to the remaining questions

**Why is `compiler-tests` scoped to `compiler/**` while the other three are repo-wide?**
Functionally it makes no difference on this run — the root-Gradle step in that job builds
only `jvmJar`s, so the repo-wide pattern would have matched the identical 21 files. The
narrower glob is defensible as precision (it states where the included build's results
live). It does have one observable side effect, visible in the table in §1: `@actions/glob`
roots the search at `<repo>/compiler`, so the compiler artifact's internal paths start at
`plugin/build/…` while the other three start at `ksrpc-*/build/…`. Anything that later
consumes all four artifacts uniformly will see that seam. Cosmetic, and I am not asking for
a change — but if the intent was precision it is worth a one-line comment saying so, and if
it was a copy that drifted, making it repo-wide would make the four steps genuinely
identical.

**Retention.** 90 days, expiring 2026-12-17. Raising, not blocking — that is ksrpc's call.
Worth knowing: to a reviewer looking at an old run, `expired: true` reads much like
`total_count: 0`, so the "did this test run" question becomes unanswerable again after the
window. 90 days comfortably outlives any review, so this is a note about archaeology on old
runs, not about the workflow.

**Category and tier.** CI infrastructure, no production code, no public API, no `.api`/
`.klib.api` movement — **tier 1**, as classified. I want to be explicit that I am agreeing
with the tier on the stronger claim rather than the weaker one: "adds a control" and "adds a
control that works" are different claims, and only the second justifies waving this through.
The second is what the artifact table in §1 shows.

### What I ran

- Fetched `c272837a` into a dedicated worktree; parsed all four workflow YAMLs with a YAML
  parser rather than grepping them, since a text scan cannot tell a step from a comment
  about a step — and this diff is 60% comments.
- Waited for run `35308813400` to complete; listed, downloaded, unzipped and XML-parsed all
  four artifacts.
- Pulled all five job logs and cross-checked the upload counts against the downloaded files.
- Built a standalone `@actions/glob` harness with assertions and a must-differ control, and
  confirmed it discriminates before trusting its "matches" results.
- Read `if-no-files-found`'s implementation in `actions/upload-artifact` at `ref=v4`.

Nothing was modified in the repo; no files committed.

### Merge

**I am deliberately not merging this.** ksrpc is not my repo and merge authority is not
mine — a clean verdict here is an input to your merge decision, not a substitute for it.
The absence of a merge is intentional, not an oversight.

### Follow-up I would suggest ksrpc file (describing, not filing)

1. **`if-no-files-found: warn` → `error` on all four steps** (§4), so an empty glob fails
   the job instead of warning into a green run. If it does not land in this PR, it is the
   one follow-up I would actually open.
2. **A cheap floor on test count.** `error` catches *zero* files; it cannot catch "148 of
   150 suites silently stopped being produced". A per-job assertion that the parsed XML
   totals at least N tests would catch partial erosion, which is the more likely regression
   now that the XML is being kept and can be counted. This PR is the prerequisite for that
   check, which is a fair reason to land it and iterate.
3. **Optional: surface the counts in the run summary.** Now that the XML is retained, a step
   writing "N tests, M failures" per job to `$GITHUB_STEP_SUMMARY` would answer "did this
   test actually run" without downloading an artifact — which is the ergonomic half of the
   problem #301 identifies.

